PR #10, opened 2015-08-27
[IMP] Surveys page removed from partner form

JavierIniesta:
There is too many pages in partner form. With smartbutton is enough.

### DIFF
--- a/partner_survey/README.rst
+++ b/partner_survey/README.rst
@@ -4,8 +4,7 @@
 Partner Survey
 ==============
 
-This module adds a smartbutton linked with surveys and "Surveys" page to
-partner form.
+This module adds a smartbutton linked with surveys.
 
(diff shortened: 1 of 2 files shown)


rafaelbn:
Dear @Ehtaga @pedrobaeza , after reviewing the number of tabs in partner form we realize that It has no sense to add a tab when you have a smartbutton. Moreover, the partner form could be very crowed of tab so we decide to remove the tab of surveys in partner form. Please review and if agree give me a feedback. Thanks!
For this PR :+1: tested in production enviroment and in runbot.

Ehtaga:
I agree, a smart button is enough.
Code review: :+1:

janetrp:
I've tried the survey and It works fine. The survey page has been removed and the smart button reflects the number of surveys of the partner. :+1:

PR #24, opened 2017-08-17
[ADD] survey answer once

thomaspaulb:
This module allows to configure surveys to forbids user answering more than once.

### DIFF
--- a/survey_answer_once/__openerp__.py
+++ b/survey_answer_once/__openerp__.py
@@ -1,0 +1,15 @@
+# -*- coding: utf-8 -*-
+{
+    'name': "Survey Answer Once",
+    'version': '8.0.1.0.0',
+    'summary': 'Allows configuring surveys so that users may only answer once',
+    'author': 'Sunflower IT,Therp BV,Odoo Community Association (OCA)',
+    'website': 'https://odoo-community.org/',
+    'license': 'AGPL-3',
+    'category': 'Marketing',
+    'depends': ['survey'],
+    'data': [
+        'views/survey_templates.xml',
+        'views/survey_view.xml',
+    ]
+}
--- a/survey_answer_once/controllers/website_survey.py
+++ b/survey_answer_once/controllers/website_survey.py
@@ -1,0 +1,38 @@
+# -*- coding: utf-8 -*-
+# © 2017 Sunflower IT (http://sunflowerweb.nl)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+import json
+import openerp.addons.survey.controllers.main as main
+from openerp.addons.web import http
+from openerp.exceptions import ValidationError
+
+
+class WebsiteSurveyExtension(main.WebsiteSurvey):
+
+    @http.route(['/survey/already.answered', '/survey/already-answered'],
+                type='http', methods=['GET'], auth="public", website=True)
+    def already_answered(self, **kwargs):
+        return http.request.website.render(
+            "survey_answer_once.duplicatesurvey")
+
+    @http.route(['/survey/disallow.invite', '/survey/disallow-invite'],
+                type='http', methods=['GET'], auth="public", website=True)
+    def disallow_invite(self, **kwargs):
+        return http.request.website.render(
+            "survey_answer_once.disallowinvite")
+
+    @http.route(['/survey/submit/<model("survey.survey"):survey>'],
+                type='http', methods=['POST'], auth='public', website=True)
+    def submit(self, survey, **post):
+        try:
+            return super(WebsiteSurveyExtension, self).submit(survey, **post)
+        except ValidationError as e:
+            if len(e.args) == 2:
+                if e.args[1] == 'duplicate_answer':
+                    return json.dumps(dict(
+                        redirect='/survey/already-answered'))
+                if e.args[1] == 'disallow_invite':
+                    return json.dumps(dict(
+                        redirect='/survey/disallow-invite'))
+            raise
--- a/survey_answer_once/models/survey_survey.py
+++ b/survey_answer_once/models/survey_survey.py
@@ -1,0 +1,64 @@
+# -*- coding: utf-8 -*-
+import logging
+
+from openerp import SUPERUSER_ID, fields, models, api, _
+from openerp.exceptions import ValidationError
+
+_logger = logging.getLogger(__name__)
+
+
+class Survey(models.Model):
+    _inherit = 'survey.survey'
+
+    allow_duplicates = fields.Boolean(
+        "Allow users to give multiple Responses",
+        default=True
+    )
+
+
+class SurveyUserInputLine(models.Model):
+    _inherit = 'survey.user_input_line'
+
+    @api.model
+    def create(self, vals):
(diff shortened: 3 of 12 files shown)


survey_answer_once/__openerp__.py
```diff
@@ -0,0 +1,17 @@
+# -*- coding: utf-8 -*-
+{
+    'name': "Survey Answer Once",
+    'version': '8.0.1.0.0',
+    'summary': 'Allows configuring surveys so that users may only answer once',
+    'description': """
+        Allows configuring surveys so that users may only answer once
+    """,
```
hbrunn: this can go away when you have a readme

survey_answer_once/__openerp__.py
```diff
@@ -0,0 +1,17 @@
...
+    'author': "Sunflower IT",
```
hbrunn: add oca

survey_answer_once/controllers/website_survey.py
```diff
@@ -0,0 +1,27 @@
...
+
+    @http.route(['/survey/submit/<model("survey.survey"):survey>'],
+                type='http', methods=['POST'], auth='public', website=True)
+    def submit(self, survey, **post):
+        try:
+            return super(WebsiteSurveyExtension, self).submit(survey, **post)
+        except ValidationError as e:
+            if len(e.args) <= 1 or e.args[1] != 'duplicate_answer':
+                raise
+            return json.dumps(dict(redirect='/survey/already-answered'))
```
hbrunn: do you really need your own controller here? Why not simply rendering the template you use above? This also has the added benefit that you can stick the current survey into the evaluation context in a convenient way

survey_answer_once/models/survey_survey.py
```diff
@@ -0,0 +1,29 @@
...
+
+class SurveyUserInputLine(models.Model):
+    _inherit = 'survey.user_input_line'
+
+    @api.multi
+    @api.constrains('user_input_id')
+    def _check_allow_duplicates(self):
+        for record in self:
+            if record.survey_id and not record.survey_id.allow_duplicates:
+                existing = self.env['survey.user_input_line'].search(
```
hbrunn: I think this is quite hard to read. My personal preference for this kind of constructs would be

```python
existing = self.env['survey.user_input_line'].search([
    ('id', '!=', record.id),
    ('question_id', '=', record.question_id.id),
]).filtered(
    lambda a: a.user_input_id.create_uid.id == self.env.uid
)
```

The beauty of this is that this is (nearly) only one instruction per line, which makes it more debug-friendly, and with this formatting, most changes will be a one line diff

hbrunn: also leave out implicit `AND`s

survey_answer_once/models/survey_survey.py
```diff
@@ -0,0 +1,29 @@
...
+                    ['&', ('id', '!=', record.id),
+                     ('question_id', '=', record.question_id.id)]).filtered(
+                    lambda a: a.user_input_id.create_uid.id == self.env.uid)
```
hbrunn: why do you need to do this on the client side anyways?

hbrunn (changes requested):
some points, and [travis](https://travis-ci.org/OCA/survey/jobs/265539616) of course

survey_answer_once/controllers/website_survey.py
thomaspaulb: We tried that first, but it didn't work, and my understanding of that was is that the `submit` controller method we are inheriting here, is used in an AJAX fashion only, so returning rendered HTML here results in an error. Is there a way around that?

survey_answer_once/models/survey_survey.py
thomaspaulb: Isn't this on the server side?

survey_answer_once/controllers/website_survey.py
hbrunn: ah, for ajax, this doesn't really work differently (you could error out here and have the client side react on this, but that's the same in green as what you do here)

survey_answer_once/models/survey_survey.py
hbrunn: now with `server` I mean postgres, and with `client` odoo. So rephrasing the question: Why can't you do the filtering as part of the search?

thomaspaulb: Well, somehow my brain thinks that `search` domain leafs don't accept dot notation to follow relations, but I am guessing I am wrong about that?

hbrunn: yes. `[('user_input_id.create_uid', '=', 42)]` is a perfectly fine domain

thomaspaulb:
@hbrunn Have I successfully responded to your review?

hbrunn:
@thomaspaulb yes, thanks. Now just add a small test for the constraint (just in the ORM, no web stuff needed imho) to make codecov happy

thomaspaulb:
@hbrunn Have added the unit test.

hbrunn:
@thomaspaulb and introduced new lint issues: https://travis-ci.org/OCA/survey/jobs/268277055#L628-L632

hbrunn (approved):

thomaspaulb:
@hbrunn There were still some issues with this, I think I have now addressed them all.

- For some reason when a user is invited personally using a token, and he follows the token link _when he is already logged in_, `self.env.uid` is set to admin (maybe because of a conflict with auth_admin_passkey? I haven't figured it out yet). This makes almost all methods to check for previously posted answers unreliable. The only reliable way I have found is to look for `user_input_id.partner_id` and when this is empty (invite was not linked to a user but just an email address), disallow him using this invite - as we cannot reliably check for duplicates - and show him a screen to ask for a new invite.
- The constraint on `survey.user.input.line` was not watertight, even though ValidationError was raised, the line record kept getting created and thus duplicate answers came into the database. I have now used a `create` hook, which works.

hbrunn:
first is caused by https://github.com/OCA/OCB/blob/8.0/addons/survey/controllers/main.py#L275 - so *all* invitations via links will suffer from this problem. But then this module is supposed to allow a use case where people can only answer suveys once, so you could suppress answers without token? This would also be consistent with the not-yet-implemented version of this for unauthorized users

second: are you sure your constraint is actually evaluated where you expect it to be evaluated? Constraints not working is something we absolutely need to fix, but is it possible that order of creates/writes play a role here? You'd have to constrain `user_input_id.survey_id.allow_duplicates` anyways in order to disallow moving some user input so some other survey where this is set and the input would add a duplicate answer.

hbrunn:
...for the first, you could change the constraint to only be effective when the user input doesn't have a token

rafaelbn:
Hi @thomaspaulb , are you interested in going ahead with this module for merging?

thomaspaulb:
@rafaelbn Yes, I could do that,

@hbrunn Were your comments binding reviews or just suggestions?

hbrunn:
they were replies to your points above, I already gave a positive review

thomaspaulb:
@rafaelbn Reviewing seems to be finished then; Travis issues seem unrelated

PR #28, opened 2018-09-27
[10.0][ADD] - Survey five stars question type

sbejaoui:
first commit for Survey five stars question type


survey_five_stars_question_type/README.rst
```diff
@@ -0,0 +1,60 @@
...
+
+.. repo_id is available in https://github.com/OCA/maintainer-tools/blob/master/tools/repos_with_ids.txt
+.. branch is "8.0" for example
+
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/{project_repo}/issues>`_. In case of trouble, please
```
benwillig: s/project_repo/survey

survey_five_stars_question_type/README.rst
```diff
@@ -0,0 +1,60 @@
...
+
+
+
+Usage
+=====
+
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/{repo_id}/{branch}
```
benwillig: s/repo_id/200
s/branch/10.0

survey_five_stars_question_type/__init__.py
```diff
@@ -0,0 +1,2 @@
+from . import models
+from . import tests
```
benwillig: Missing blank line at the end of file

survey_five_stars_question_type/models/survey_user_input_line.py
```diff
@@ -0,0 +1,34 @@
+# -*- coding: utf-8 -*-
+# Copyright 2018 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, models
+
+
+class SurveyUser_input_line(models.Model):
```
benwillig: s/SurveyUser_input_line/SurveyUserInputLine

survey_five_stars_question_type/models/survey_user_input_line.py
```diff
@@ -0,0 +1,34 @@
...
+
+    _inherit = 'survey.user_input_line'
+
```
benwillig: Shouldn't we add a python constrains which check that the answer is between 1 and 5?

survey_five_stars_question_type/tests/test_survey.py
```diff
@@ -0,0 +1,28 @@
...
+            ('aaa', 'This is not a number'),
+            ('0', 'Answer is not in the right range'),
+            ('6', 'Answer is not in the right range')]
+        msg = "Validation function for type numerical_box is unable to " \
+              "notify if answer is violating the validation rules"
+        for i in range(len(results)):
+            self.assertEqual(question.validate_question(
+                {answer_tag: results[i][0]}, answer_tag),
+                {answer_tag: results[i][1]},
+                msg=msg)
```
benwillig: Shouldn't we check that there is no error when providing a valid value?

survey_five_stars_question_type/views/survey_template.xml
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <data>
```
benwillig: data tag is deprecated

benwillig (changes requested):
Some minor comments.

Could you split the readme.rst in separate files like [here](https://github.com/OCA/mis-builder/tree/10.0/mis_builder/readme). The readme file will be generated automatically after the PR is merged.

survey_five_stars_question_type/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Survey five stars question type',
+    'summary': """
+        This module add five stars rating as question type for survey page""",
+    'version': '10.0.1.0.0',
+    'license': 'AGPL-3',
+    'author': 'ACSONE SA/NV,Odoo Community Association (OCA)',
+    'website': 'http://acsone.eu',
```
benwillig: The website should be the OCA repository url.

pedrobaeza:
Please avoid plurals on the name, and following usual conventions, the name should be `survey_question_type_five_star`.

pedrobaeza:
Check also Travis status

survey_five_stars_question_type/models/survey_user_input_line.py
sbejaoui: Already done at `models/survey_question.py: validate_star_rate`

benwillig: I'm talking about a python model constrain, because if I do a simple ```write``` with an invalid value, there will be no error.

sbejaoui: Yes I agree, a simple `write` bypass the constraint, we may have a question type matrix and as answer we put text.
But it's done like that in survey module, the check is done only at user submit.
A method `save_{question_type}` call `validate_{question_type}`.

benwillig: Odoo doesn't use pylint/flake8 but we do :smile:
We might need more opinion about that but I would like to have a constrain for that. I would say, it isn't because Odoo did it wrong that we should follow the same path ^^.

benwillig (changes requested):
After some functionnal testing, I found some points which need to be fixed

1) When answering to a survey with the new question type, if you don't click on any stars and you submit the form. There will be an error. This error is raised because the value for the answer is blank and isn't between 1 and 5. So I think,we should allow people to mark the answer without any star, with a 0 value then.

2) After answering the survey, you can check the answer you gave. The new question type is not handled in this view.

3) If you want to check the survey results with the "View results" button available in the survey form view. The new question type is not handled here so we can't have any statistics.

survey_question_type_five_star/static/src/js/survey.js
```diff
@@ -1,13 +1,44 @@
...
+    $(document).ready(function() {
+        //Print Result case
+        var the_form = $('.js_surveyform');
+        var prefill_controller = the_form.attr("data-prefill");
+        if (!the_form.length) {
+            return $.Deferred().reject("DOM doesn't contain '.js_surveyform'");
+        }
+        function prefill() {
+            if (!_.isUndefined(prefill_controller)) {
+                var prefill_def = $.ajax(prefill_controller, {
```
benwillig, line 15: This call is already done in the base module. Can't we just add a onchange listener on the input? Then when the base module updates inputs value, you can determine which star should be highlighted. I think we could use the ```Widget``` class, this widget would take a DOM element/JS selector in parameter. And the target object would be the HTML structure of the "Five star rate" representation. When the widget is instanciated, it would check the current value of the input to highlight stars, and it would add an onchange listener to handle the possible changes. The widget should be readonly for the result view and editable when answering.

survey_question_type_five_star/static/src/js/survey.js
```diff
@@ -1,13 +1,44 @@
...
+                        dataType: "json"
+                    })
+                    .done(function(json_data) {
+                        _.each(json_data, function(value, key) {
+                            var input = the_form.find(".form-control[name=" + key + "]");
+                            $(input).parent().find("label").each(function(label_index) {
+                                if ((5 - label_index) <= value) {
+                                    $(this).addClass("checked");
```
benwillig: We have two different logics to highlight stars, so I think a widget would be better here.

benwillig (changes requested):

survey_question_type_five_star/models/survey_survey.py
```diff
@@ -0,0 +1,44 @@
...
+                ):
+                    all_inputs.append(input_line.value_number)
+                    result_summary["input_lines"].append(input_line)
+            if all_inputs:
+                result_summary.update(
+                    {
+                        "average": round(sum(all_inputs) / len(all_inputs), 2),
+                        "max": round(max(all_inputs), 2),
+                        "min": round(min(all_inputs), 2),
+                        "sum": sum(all_inputs),
```
tarteo, line 37: If this is stored in a variable (like sum_all_inputs) and in the 2 cases it's more efficient. But the gain is minimal so it's not blocking :P

survey_question_type_five_star/static/src/js/survey.js
```diff
@@ -0,0 +1,44 @@
...
+        $(".rate > label").click(function() {
+            var label_items = $(this).parent().find("label");
+            var value = label_items.length - $(this).index();
+            label_items.removeClass("checked");
+            label_items.slice($(this).index()).addClass("checked");
+            $(this).parent().find("input").val(value);
+        });
+    });
+
+});
```
tarteo: New line

tarteo (changes requested):
Cool module! Code looks good, I just have a few minor comments. 
When I make a question mandatory, I still can skip it. 
It would be nice be to be able to select the amount of stars for a question.

sbidoul:
Hi all, thanks for the review. We have limited budget to finalize this.

I see one important issue: "When I make a question mandatory, I still can skip it.". @sbejaoui is that easy to fix?

@tarteo Making the number of stars configurable looks too complex for our budget. May be we can rename the module to survey_question_type_star, and leave the configurability for later.

sbidoul:
[updated] module name in previous comment

sbidoul:
@benwillig I understand your comments are nice to have refactorings. Are they blocking for you?

sbejaoui:
@sbidoul , @tarteo ,
For the mandatory option, I think we should decide if we consider 0 as a valid answer or not.
When user skip a question and any of the five stars is highlighted a 0 value is sent.
In this case the mandatory option make no sens.
So I propose to consider valid answer for mandatory question from 1 to 5.

tarteo (approved):
@sbejaoui @sbidoul I tested it again and it works perfectly. Thanks. Do you have budget to fix the jslint errors?

sbejaoui:
@tarteo , I pushed a fix for the jslint

rafaelbn (approved):

sbejaoui:
Hi @rafaelbn ,

I believe this can be merged

PR #33, opened 2020-05-05
[13.0][MIG] partner_survey: Migration to 13.0

skeller1:
partner_survey

**Idea**

Odoo 13.0 introduced the certification count stats button on partner form. This is already implemented as a survey relation with filter on quiz option=true.
Used the same implementation for survey counting.
If the survey count is zero or equal to the certification count the survey button is hidden.

- [x] added simple tests
- [x] Refactor


bruno-zanotti:
@skeller1 
I've tested and everything works properly.
About this "Understand why attrs invisible with ('survey_count', '=','certification_count') domain is not working"
It is not working because the third value of the triple is just a value, so you are asking if the field 'survey_count' conteins the string "certification_count".
I think you are solving this problem in a good way.

bruno-zanotti:
Maybe you should announce that you are working on this in the issue #31

skeller1:
> @skeller1
> I've tested and everything works properly.
> About this "Understand why attrs invisible with ('survey_count', '=','certification_count') domain is not working"
> It is not working because the third value of the triple is just a value, so you are asking if the field 'survey_count' conteins the string "certification_count".
> I think you are solving this problem in a good way.

Yeah, learned it the hard way:
 model field on the left side, static value on the right side for proper condition definition in xml.
So this is not possible: `   ('survey_count', '=','certification_count')`
Thank you for clarifying

bruno-zanotti:
Hi @skeller1, thanks for your job!
Is there something missing in the pr or could be merged?
Maybe you could ping to the po of OCA asking for a review.

skeller1:
> Hi @skeller1, thanks for your job!
> Is there something missing in the pr or could be merged?
> Maybe you could ping to the po of OCA asking for a review.
I think the addon has the same functionality as V12.
I would like to add an survey generation from contacts list view (custom wizard). Maybe this is a complete new addon.

But so far: I think we could merge this for V13. 
AFAIK maintainer will merge as soon as possible. It tooks some time ......

Thank you

pedrobaeza:
There seems to not be interest in reviewing this module so far, but trusting on CI and the contributor, I merge it:

/ocabot merge nobump

Note that this is the exception, not the rule. @Bruno-Zanotti you can review it as well: https://odoo-community.org/page/review, and you can ask for reviews in exchange of you reviewing other PRs.

skeller1:
> Note that this is the exception, not the rule.

@pedrobaeza Thank you for merging. No offense meant. Thank you for all your work.

PR #36, opened 2020-06-16
[12.0][ADD] survey_conditional_question module

remi-filament:
Add module for conditional question (depends on survey_backend_readability - PR #35 )


survey_conditional_question/__manifest__.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2020 Le Filament (<https://www.le-filament.com>)
+# License GPL-3.0 or later (https://www.gnu.org/licenses/gpl).
+
+{
+    "name": "Survey Conditional Question",
+    "summary": "Display question depending on answer to previous one",
+    "version": "12.0.1.0.0",
+    "development_status": "Beta",
+    "website": "https://github.com/OCA/survey",
+    "author": "Le Filament, Vuente, Odoo Community Association (OCA)",
```
StefanRijnhart, line 10: Are you the author of the Vuente code, or has Vuente signed the OCA CLA?

remi-filament: Hi @StefanRijnhart, no I am not the author, yet we inspired our module (mainly the field names and the logic of the template XML file) from one found on apps.odoo.com from Vuente (https://apps.odoo.com/apps/modules/9.0/vuente_survey_extra/) so we kept them in author.
I have been trying to figure out how to reach out for them to ask whether 
- they have signed the CLA or not
- they want their name as author of our module
- they are OK with this module being pushed to OCA

But I have not found any way to get in touch with them. What should we do in that case ?

rafaelbn: Hi @remi-filament , you can write yo legal@odoo-community.org to check what they say about or you can ask to push OCA this one: https://github.com/ingadhoc/survey/tree/11.0/survey_conditional_questions

remi-filament: Thanks @rafaelbn , will do
The other module you refer to is not relevant for our use case since conditional questions are related to questions from another page, you cannot configure for exemple that if you answer A to question 1, question 2 will be displayed on the same page.

etobella: I checked ingadhoc survey system and it is different, it hides according to previous pages, not the current ones.
In any case, as I can see from @remi-filament comments, the code was inspired from Vuente, no taken from there, so I don't see if there is any issues.

StefanRijnhart: Not sure if this helps, but here is my v10 branch that adds same page conditional questions to the adhoc version of the module. The ingadhoc people agreed that I put this in the OCA but because it was an old version I never did. https://github.com/ewac-nl/adhoc-survey/tree/imp/10.0%2Fsurvey_conditional_questions%2Fsame_page

PR #40, opened 2021-04-27
[ADD] survey_resource_booking: link resource bookings and surveys

yajo:
This module extends the functionality of surveys and resource bookings
extensions to support linking resource booking types to surveys and resource
bookings  read and request answers.

Draft until merged:

- [x] https://github.com/OCA/calendar/pull/25

@Tecnativa TT28205 TT28203


survey_resource_booking/models/resource_booking.py
```diff
@@ -0,0 +1,59 @@
...
+                    % one.display_name
+                )
+            if one.survey_user_input_id.state == "done":
+                raise UserError(
+                    _(
+                        "This booking's requester was already invited to fill survey: %s"
+                    )
+                    % one.display_name
+                )
+            if not one.survey_user_input_id:
```
joao-p-marques: Shouldn't the first check be after the second one? What if `survey_user_input_id` does not exist? 
```suggestion
            if not one.survey_user_input_id:
                # ....
            if one.survey_user_input_id.state == "done":
                raise UserError(
                    _(
                          "This booking's requester was already invited to fill survey: %s"
                      )
                      % one.display_name
                 )
```

joao-p-marques (commented):
Code review :+1: 

Just a small comment. Also, Travis is red (apart from the missing module, there is a flake8 error)

survey_resource_booking/models/resource_booking.py
yajo: Well, it seems misleading but it's OK.

In Odoo, when you access a zero-recordset field, it will always be `False` (or an empty recordset, if the field is relational).

survey_resource_booking/models/resource_booking.py
```diff
@@ -0,0 +1,68 @@
...
+from odoo.exceptions import UserError
+
+
+class ResourceBooking(models.Model):
+    _inherit = "resource.booking"
+
+    survey_user_input_id = fields.Many2one(
+        "survey.user_input",
+        string="Survey user input",
+        track_visibility="onchange",
```
pedrobaeza, line 16: Why not being this field `readonly`? You can always fix any mess through shell, mass editing...

pedrobaeza (approved):

survey_resource_booking/models/resource_booking.py
yajo: It's because, if for any reason you want to invite the requester to fill the survey again (invalid input? lost invitation? etc.), you have to empty this field before. Otherwise, you'll get an error telling you that he was already invited.

It could have been done by other means, but at least that's how it is implemented.

pedrobaeza: That's a bit ugly and the invite method should handle it in my opinion. Isn't a code like this enough:
```python
   if self.survey_user_input_id:
       self.survey_user_input_id = False
```

joao-p-marques (approved):

survey_resource_booking/models/resource_booking.py
pedrobaeza: What do you think about this? At least a reference in the ROADMAP?

yajo: It's just a different way of supporting something that is right now supported. The customer never complained about this, so I'd just leave things as they are.

PR #51, opened 2022-01-25
[13.0][MIG] survey_resource_booking: Migration to 13.0

victoralmau:
Migration to 13.0

Please @chienandalu and @pedrobaeza can you review it?

@Tecnativa TT31325


survey_resource_booking/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,3 @@
+* `Tecnativa <https://www.tecnativa.com>`_:
+
+  * Jairo Llopis
```
pedrobaeza, line 3: Put yourself here

test-requirements.txt
```diff
@@ -0,0 +1 @@
+cssselect
```
pedrobaeza, line 1: Why is this needed?

pedrobaeza (approved):
Tested on runboat and code review

test-requirements.txt
victoralmau: It is a dependency of the `resource_booking` addon that is only used in tests: https://github.com/OCA/calendar/blob/13.0/test-requirements.txt

chienandalu (approved):

PR #65, opened 2022-12-30
[13.0][ADD] survey_sale_generation: New module 13.0

chienandalu:
This module allows to generate quotations from surveys submits.

cc @Tecnativa TT40983


survey_sale_generation/models/survey_user_input.py
```diff
@@ -0,0 +1,79 @@
...
+        quotable_lines = self.user_input_line_ids.filtered("value_suggested.product_id")
+        quotable_lines += self.user_input_line_ids.filtered(
+            lambda x: x.question_id.product_id and not x.skipped
+        )
+        if not quotable_lines:
+            return res
+        vals = self._prepare_quotation()
+        vals["order_line"] = [
+            (0, 0, self._prepare_quotation_line(line)) for line in quotable_lines
+        ]
```
victoralmau, line 76: Maybe it would be a good idea to add lines (`order_line`) in `_prepare_quotation()` method directly. what do you think?

victoralmau (changes requested):
Good job, some comments related to functional review:

- Can you add tests to validate all steps?
- Add `sale_order_id` field in form view of `survey.user_input `for easy navigation.
- Add `survey_user_input_id` field in form view of `sale.order` for easy navigation.
- Question type: Multiple choice: only one answer: Use case is not correct (it generates as many lines as answers although only one can be selected).

survey_sale_generation/models/survey_user_input.py
chienandalu: I prefer to leave it split so other modules can hook directly in the future

gurneyalex (approved):

chienandalu:
Thanks @gurneyalex . I was hoping to add some fixes. Can you abort?

chienandalu:
Nevermind. I'll do it in an extra PR :sweat_smile:

PR #70, opened 2023-01-25
[13.0][ADD] survey_formio: add compatibility between survey and form.io

jdoutreloux:
This module adds the necessary function to generate a form.io compatible JSON that represents an odoo survey.
It also adds the function that creates or updates a survey_user_input when given the JSON returned by form.io


survey_formio/models/survey.py
```diff
@@ -0,0 +1,149 @@
...
+        """
+        Returns a string representing the form.io display
+        Correct values are form, wizard or pdf. At the moment only
+        form is supported.
+        In the future use questions_layout to determine display.
+        """
+        self.ensure_one()
+        return "form"
+
+    def _get_components(self, day_first=True):
```
sbidoul:
```suggestion
    def _get_formio_components(self, day_first=True):
```

survey_formio/models/survey_question.py
```diff
@@ -0,0 +1,216 @@
...
+        - values: list of dict {"label": label, "value": value}
+        """
+        self.ensure_one()
+        return {
+            "values": [
+                {"label": label.value, "value": label.id} for label in self.labels_ids
+            ]
+        }
+
+    def _get_formio_dict(self, day_first=True):
```
sbidoul:
```suggestion
    def _get_formio_component(self, day_first=True):
```

survey_formio/models/survey.py
```diff
@@ -0,0 +1,149 @@
...
+            components.append(question._get_formio_dict(day_first))
+        return components
+
+    def generate_formio_json(self, day_first=True):
+        """
+        Returns a form.io conform json representation of a survey.
+
+        Parameters:
+        day_first: boolean that determines if the day comes first or
+                   not in dates
```
sbidoul:
```suggestion
        day_first: True if the day input should be the first input item for date fields
```

This is a bit strange to propage this day_first options across all methods.
Is it really the only thing that is and will be configurable?
Maybe we need an options data structure?
Or be more opinionated (say, dayFirst=true always) and leave configurability for later when we have concrete use cases?

survey_formio/models/survey_question.py
```diff
@@ -0,0 +1,216 @@
...
+    "datetime": "datetime",
+    "simple_choice": "radio",
+    "multiple_choice": "selectboxes",
+}
+
+
+class SurveyQuestion(models.Model):
+    _inherit = "survey.question"
+
+    def _get_common_parameters(self):
```
sbidoul: When adding a lot of methods to an existing model, let's prefix them to avoid name conflicts and for clarify, so:

```suggestion
    def _get_formio_common_parameters(self):
```

Alternatively, put everything in a FormIOConverter abstract model with only model methods.

.pre-commit-config.yaml
```diff
@@ -15,7 +15,7 @@ exclude: |
   # You don't usually want a bot to modify your legal texts
   (LICENSE.*|COPYING.*)
 default_language_version:
-  python: python3
+  python: python3.6
```
sbidoul: Is this necessary?

survey_formio/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,8 @@
+This module adds compatibility bewtween form.io and Survey. It adds a function on survey.survey that generates
```
sbidoul:
```suggestion
This module adds compatibility bewtween `form.io <https://formio.github.io/formio.js/app/builder>`_ and Survey. It adds a function on survey.survey that generates
```

survey_formio/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,8 @@
...
+a form.io compatible JSON that represents that survey. It also adds a function to generate a survey.user_input
+from a JSON returned by form.io.
+Use generate_formio_json to get the json representing a survey.
```
sbidoul:
```suggestion
Use `survey.survey::generate_formio_json()` to get the json representing a survey that can be rendered using the form.io widget.
```

survey_formio/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,8 @@
...
+Use user_input_from_formio to create a user_input with the answer from form.io
```
sbidoul:
```suggestion
Use `survey.survey::user_input_from_formio()` to create a user_input with the answer from form.io
```

survey_formio/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,8 @@
...
+
+As of now this export does not support validation options other than survey_question.constr_mandatory. It also does
+not work with question of type 'matrix'
```
sbidoul: Move this to the roadmap section?

survey_formio/models/survey_question.py
```diff
@@ -0,0 +1,216 @@
...
+        - validate : only constr_mandatory supported at the moment,
+                     will be handled in its own function
+
+        See https://github.com/formio/formio.js/wiki/Components-JSON-Schema#common-parameters
+        for more information
+        """
+        self.ensure_one()
+        return {
+            "type": ODOO_TO_FORM_IO_TYPES_DICT[self.question_type],
+            "key": self.id,
```
sbidoul:
```suggestion
            "key": f"q{self.id},
```

It may be better to prefix these keys, so they are not integers. It's a bit unusual to have integers as field names in a form, and may break things downstream.

sbidoul (commented):
Looks promising. A few style comments. I've not looked into the logic yet.

survey_formio/models/survey.py
```diff
@@ -0,0 +1,149 @@
...
+        display = self._get_formio_display()
+        title = self.title
+        components = self._get_components(day_first)
+
+        form_json = json.dumps(
+            {"display": display, "title": title, "components": components}
+        )
+        return form_json
+
+    def user_input_from_formio(self, formio_input, user_input_id=None, day_first=True):
```
sbidoul: This function is a big long. Can it be split in small functions for each answer type?

survey_formio/models/survey.py
sbidoul: Does dayFirst influence the display of the form, or the datetime format in the answer, or both?

survey_formio/models/survey.py
```diff
@@ -0,0 +1,149 @@
...
+                    # in Python v3.7+ and Odoo v13 is compatible with Python v3.6
+                    datetime_answer = form_io_answers[str(question.id)]
+                    form_io_answers[str(question.id)] = (
+                        datetime.strptime(datetime_answer[:19], "%Y-%m-%dT%H:%M:%S")
+                        + timedelta(
+                            hours=int(datetime_answer[20:22]),
+                            minutes=int(datetime_answer[23:]),
+                        )
+                        * (-1 if datetime_answer[19] == "+" else 1)
+                    ).strftime("%Y-%m-%d %H:%M:%S")
```
sbidoul: Woud this work?

`datetime.datetime.strptime(datetime_answer, "%Y-%m-%dT%H:%M:%S%z").astimezone(pytz.UTC)`

survey_formio/models/survey.py
jdoutreloux: It influences both. Without setting it I can't know which date format I'll receive. For now I think it's the only configurable thing. 
Should I just say dayFirst=True for now then?

sbidoul: > It influences both. Without setting it I can't know which date format I'll receive. For now I think it's the only configurable thing.

That is surprising. Is there not a mechanism to tell it to return an ISO format for the dates?

> Should I just say dayFirst=True for now then?

Yes.

jdoutreloux: > That is surprising. Is there not a mechanism to tell it to return an ISO format for the dates?
If there's one I couldn't find it. It surprised me too, especially because the datetime component always returns the same format.
> Yes.
Ok. I'll do that

sbidoul: Can we use the `datetime` widget with `"enableTime": false` instead of the day widget?

survey_formio/models/survey.py
jdoutreloux: Done

survey_formio/models/survey.py
jdoutreloux: %z is not supported in Python 3.6

.pre-commit-config.yaml
jdoutreloux: I will remove this from my commit.

survey_formio/models/survey_question.py
jdoutreloux: I've added the prefix

survey_formio/models/survey_question.py
jdoutreloux: Done

survey_formio/models/survey_question.py
jdoutreloux: Done

survey_formio/readme/DESCRIPTION.rst
jdoutreloux: Done

survey_formio/readme/DESCRIPTION.rst
jdoutreloux: Done

survey_formio/readme/DESCRIPTION.rst
jdoutreloux: Done

survey_formio/readme/DESCRIPTION.rst
jdoutreloux: I've moved the info about validation. I didn't move the second part, I've only reworked it a bit. I don't think it should be in the roadmap because I don't think it'll ever be supported as there's no corresponding component in formio.

survey_formio/models/survey.py
```diff
@@ -0,0 +1,153 @@
...
+                    form_io_answers.pop(question_label)
+                    form_io_answers.update(
+                        self._from_formio_multiple_choice_to_odoo(
+                            question_label, answers_dict
+                        )
+                    )
+                self.env["survey.user_input_line"].save_lines(
+                    user_input_id.id, question, form_io_answers, question_label
+                )
+        user_input_id._mark_done()
```
sbidoul: Would it be more flexible to let the caller do `_mark_done()` ?

survey_formio/models/survey_question.py
```diff
@@ -0,0 +1,186 @@
...
+        if self.question_type == "numerical_box":
+            specific = self._get_formio_number_parameters()
+        if self.question_type == "date":
+            specific = self._get_formio_datetime_parameters(is_datetime=False)
+        if self.question_type == "datetime":
+            specific = self._get_formio_datetime_parameters()
+        if self.question_type == "simple_choice":
+            specific = self._get_formio_radio_parameters()
+        if self.question_type == "multiple_choice":
+            specific = self._get_formio_selectboxes_parameters()
```
sbidoul, line 174: I'd use `elif`'s here. Also a `else` with a warning that the question type is not supported?

survey_formio/models/survey.py
```diff
@@ -0,0 +1,152 @@
...
+
+        form_json = json.dumps(
+            {"display": display, "title": title, "components": components}
+        )
+        return form_json
+
+    def user_input_from_formio(self, formio_input, user_input_id=None):
+        """
+        If no user_input_id is given, creates a user_input with the values in formio_input.
+        If there is a user_input_id, updates that user_input with the values in formio_input.
```
marielejeune: FYI, OCA guidelines ask to use present imperative in comments: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#directories:~:text=Use%20present%20imperative%20(Fix%20formatting%2C%20Remove%20unused%20field)%20avoid%20appending%20%27s%27%20to%20verbs%3A%20Fixes%2C%20Removes

marielejeune (approved):
Great work!

rousseldenis (approved):
Code review

PR #72, opened 2023-02-06
[14.0][ADD] survey_multiple_choice_max_answer: New module

marielejeune:
This module adds the possibility to limit the number of answers for a multiple choice question.


pedrobaeza:
Thanks for the contribution.

Please don't include plurals in module name.

qgroulard (approved):

survey_multiple_choice_max_answer/models/survey_question.py
```diff
@@ -0,0 +1,50 @@
...
+    def _validate_choice(self, answer, comment):
+        res = super()._validate_choice(answer, comment)
+        if res or self.question_type != "multiple_choice":
+            return res
+
+        # When only 1 suggested answer is received from website in controller,
+        # it should be given as a list of length 1: [sugg_ans_id],
+        # but it is given as a string: str(sugg_ans_id)
+        answer_count = len(answer) if isinstance(answer, list) else 1
+        if self.validation_required and not (
```
jdoutreloux: I think you should check if validation_multiple_answers_max is different than 0. 
If your module and another that extends the validation functionality too are both installed, if the user doesn't set the values because he does not want to use them but instead leaves them at 0 (default value of an Integer) then the validation will fail.

survey_multiple_choice_max_answer/models/survey_question.py
```diff
@@ -0,0 +1,50 @@
...
+
+from odoo import fields, models
+
+
+class SurveyQuestion(models.Model):
+
+    _inherit = "survey.question"
+
+    validation_multiple_answers_min = fields.Integer(
+        "Minimum Number of Answers", default=0
```
jdoutreloux: The default value of an integer is 0. You don't have to specify it

survey_multiple_choice_max_answer/models/survey_question.py
marielejeune: You're right. Changed + unit test added

survey_multiple_choice_max_answer/models/survey_question.py
marielejeune: Right, changed

jdoutreloux (approved):

PR #74, opened 2023-03-07
[13.0][ADD] survey_mail_result: New module

chienandalu:
cc @Tecnativa TT40981


survey_result_mail/README.rst
```diff
@@ -0,0 +1,107 @@
...
+    :target: https://translation.odoo-community.org/projects/survey-13-0/survey-13-0-survey_result_mail
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/200/13.0
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+This module extends the functionality of surveys to allow you to send an automatic
+email with the survey answers to the customer as soon as the questionary is completed.
```
pedrobaeza, line 29:
```suggestion
email with the survey answers to the filler as soon as the questionary is completed.
```

survey_result_mail/models/survey_survey.py
```diff
@@ -0,0 +1,12 @@
+# Copyright 2023 Tecnativa - David Vidal
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import fields, models
+
+
+class SurveySurvey(models.Model):
+    _inherit = "survey.survey"
+
+    send_survey_answers = fields.Boolean(
```
pedrobaeza: Avoid plurals in names:

```suggestion
    send_result_mail = fields.Boolean(
```

survey_result_mail/views/survey_survey_views.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+    <record id="survey_form" model="ir.ui.view">
+        <field name="model">survey.survey</field>
+        <field name="inherit_id" ref="survey.survey_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='question_and_page_ids']/.." position="inside">
+                <field name="send_survey_answers" />
```
pedrobaeza: You need to put a group to show the label:

```suggestion
                <group>
                    <field name="send_survey_answers" />
                </group>
```

Selección_069

Maybe this can go in the "Thank you message" page?

survey_result_mail/reports/survey_answer_report.xml
```diff
@@ -0,0 +1,45 @@
...
+                    t-value="user_input.with_context(lang=user_input.partner_id.lang or user_input.env.lang)"
+                />
+                <t
+                    t-call="survey_result_mail.survey_result_report_view_document"
+                    t-lang="user_input.partner_id.lang or user_input.env.lang"
+                />
+            </t>
+        </t>
+    </template>
+    <report
```
pedrobaeza, line 35: Why putting this as a report. The idea is to have the answers in the mail body itself.

survey_result_mail/models/survey_user_input.py
```diff
@@ -0,0 +1,48 @@
...
+        )
+
+    def _compute_survey_result(self):
+        for user_input in self:
+            user_input.survey_result = user_input._render_user_input()
+
+    def _mark_done(self):
+        """Send the answers when submitted on the so configured surveys"""
+        res = super()._mark_done()
+        for user_input in self.filtered("survey_id.send_survey_answers"):
```
pedrobaeza: We are only having the email address if the user is logged in. Isn't it? So maybe we should restrict it for now or have a way to indicate the email source (for example, one of the answers).

In fact, pressing "Test", this happens:

Selección_070

pedrobaeza (changes requested):
I haven't been able to test more, as sending invitations seems to not work, not sure if it's something due to this module.

survey_result_mail/reports/survey_answer_report.xml
chienandalu: The aim is to be able to reuse the base results as much as possible and avoid falling into a jinja template hell (we don't have bootstrap, for example)

survey_result_mail/models/survey_user_input.py
chienandalu: Right, I should fix that.

survey_result_mail/reports/survey_answer_report.xml
chienandalu: I explored how mass_mailing is doing it, but the conversion from bootstrap is made by the js of the editor, so we can't use that either.

pedrobaeza: You can render a QWeb template as HTML, and inject it in the mail body.

chienandalu: It would be without any style.

pedrobaeza: Yes, but it's OK in this version, In v15, you can put styles in header.

chienandalu: Maybe a minimized and cutoff version of bootstrap would do in that case... although we'll have to deal with mail client css limitations...

pedrobaeza: Yes, but for now the most important thing is to have the content inside the mail, even in plain text, but it's not the case, as you can apply at least bold and so on.

survey_result_mail/views/survey_survey_views.xml
chienandalu: > Maybe this can go in the "Thank you message" page?

What do you mean?

pedrobaeza: Move the field to the `<page>` in the notebook called "Thank you message"

chienandalu: Looks like that view item doesn't exist anymore. I put it in the Questions section of the Options tab.

survey_result_mail/models/survey_survey.py
chienandalu: done

survey_result_mail/models/survey_user_input.py
chienandalu: done

survey_result_mail/README.rst
chienandalu: fixed

chienandalu:
Ready for review again

pedrobaeza (approved):
Can you please squash?

chienandalu:
Done. Isn't it possible to squash a PR commits via ocabot?

pedrobaeza:
> Done. Isn't it possible to squash a PR commits via ocabot?

I'm afraid not.

/ocabot merge nobump

PR #76, opened 2023-05-12
Fix missing dependency

alessandro-fiorino:
Installing survery_question_type_five_start without survey_question_type_nps leads to errors when opening the survey


chienandalu:
Check https://github.com/OCA/survey/pull/114

PR #82, opened 2023-06-16
[14.0][MIG] survey_crm_generation: migration to 14.0

Rferri44-S73:
Standard migration to 14.0


chienandalu (commented):
You can include https://github.com/OCA/survey/pull/100

PR #84, opened 2023-06-20
[13.0][IMP] survey_contact_generation: multiple choice

chienandalu:
- Added support for multiple choice questions
- We can choose many2one field values in multiple choice single
- We can use many2many field values in multiple choice multiple
- We can use whatever field in multiple choices
- Now we can use the comment field as a container for any question info that doesn't fit into other fields.

cc @Tecnativa TT40985

please review @stefan-tecnativa

### DIFF
--- a/survey_contact_generation/models/survey_question.py
+++ b/survey_contact_generation/models/survey_question.py
@@ -18,11 +18,13 @@ class SurveyQuestion(models.Model):
     @api.depends("question_type")
     def _compute_allowed_field_ids(self):
         type_mapping = {
-            "textbox": ["char"],
+            "textbox": ["char", "text"],
             "free_text": ["text"],
-            "numerical_box": ["integer", "float"],
-            "date": ["date"],
-            "datetime": ["datetime"],
+            "numerical_box": ["integer", "float", "text", "char"],
+            "date": ["date", "text", "char"],
+            "datetime": ["datetime", "text", "char"],
+            "simple_choice": ["many2one", "text", "char"],
+            "multiple_choice": ["many2many", "text", "char"],
         }
         for record in self:
             record.allowed_field_ids = (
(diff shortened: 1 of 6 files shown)


pedrobaeza:
@chienandalu Please check tests

survey_contact_generation/models/survey_question.py
```diff
@@ -18,11 +18,13 @@ class SurveyQuestion(models.Model):
...
+            "textbox": ["char", "text"],
             "free_text": ["text"],
-            "numerical_box": ["integer", "float"],
-            "date": ["date"],
-            "datetime": ["datetime"],
+            "numerical_box": ["integer", "float", "text", "char"],
+            "date": ["date", "text", "char"],
+            "datetime": ["datetime", "text", "char"],
+            "simple_choice": ["many2one", "text", "char"],
+            "multiple_choice": ["many2many", "text", "char"],
```
stefan-tecnativa, line 27: When I try to use a multiple_choice question on a char field, i select in the survey more than one option but in the contact the field is filled with only one of the answers. Isn't this the same behaviour as the simple_choice?

What i was expecting is some kind of string concatenation between the answers selected. Doe this make any sense? Should be reconsider the multiple_choice questions possibilities?

chienandalu: Good catch, it should return all the responses. Can you add the fix for it?

stefan-tecnativa: Done, please review it.

chienandalu (commented):
Thanks @stefan-tecnativa Please fixup your last test commit into your first one:

image

chienandalu (commented):
Thanks :)

Please review @victoralmau @pedrobaeza

pedrobaeza (approved):
/ocabot merge minor

chienandalu:
Forward port it to the v15 migration, please

PR #89, opened 2023-07-14
[14.0][ADD] survey_question_type_binary

zamberjo:
New module that enables the option to upload files in the surveys, controlling both the type of file uploaded and its size.
It also adds the possibility of adding several files with the option "Multiple: Binary".


zamberjo:
Pre-commit is fixed with https://github.com/OCA/survey/pull/90.

belenaures (approved):
Checked to see if I could upload correctly more than 1 image in a binary question, works ok

chienandalu (approved):
Tested :+1:

PR #94, opened 2023-08-21
[15.0][ADD] survey_answer_linked_record: New module

chienandalu:
cc @Tecnativa TT40984


chienandalu:
Closing as it was just experimental for customer requisite and it isn't needed at this moment

PR #95, opened 2023-08-22
[MIG] survey_legal: Migration to 15.0

chienandalu:
Aside from migration adaptations:

- Drop website_require_legal dependency
- Validation is now performed in js insead of the view so we need to do so.
- Additional control for the legal terms link and text

cc @Tecnativa TT44775

please review @victoralmau @pilarvargas-tecnativa


victoralmau (approved):
Code and functional review OK

rafaelbn (approved):
👍🏼 

Tested in runbot 

imagen

chienandalu:
Thanks @rafaelbn :wink: anyway, the merge command wasn't `patch`, but `nobump` as the version is already raised in the migration process

rafaelbn:
I know, there are two ways 😄 . @rousseldenis and me use `patch` always. Reasons are somewhere in a detailed explanation in a PR.

pedrobaeza:
But that "reason" was for older versions (13-). On recent versions, no version bumping is needed, as @rousseldenis can confirm.

PR #96, opened 2023-08-22
[15.0][ADD] survey_placeholder: New module

chienandalu:
Basically a backport of https://github.com/odoo/odoo/pull/78682 from v16

cc @Tecnativa TT40984


survey_placeholder/readme/ROADMAP.rst
```diff
@@ -0,0 +1 @@
+In v16 the feature is already included in core.
```
pedrobaeza:
```suggestion
* In v16, the feature is already included in core.
```

pedrobaeza (approved):

survey_placeholder/readme/ROADMAP.rst
chienandalu: Done

survey_placeholder/views/survey_question_views.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="survey_question_form" model="ir.ui.view">
+        <field name="model">survey.question</field>
+        <field name="inherit_id" ref="survey.survey_question_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//div[hasclass('oe_title')]" position="after">
```
stefan-tecnativa, line 7: In 15 this class is applied in the title section only, so the field is being placed under the questions name, not inside the options tab: [](https://github.com/odoo/odoo/blob/15.0/addons/survey/views/survey_question_views.xml#L16C64-L16C64)

chienandalu: Yes, the options in the question form are different in v16. So I decided to put it close to the question title which is quite related to the placeholder

stefan-tecnativa: Ah, undestood. It looks correct there for me.

stefan-tecnativa (approved):

PR #97, opened 2023-09-28
[13.0] [FIX] survey_formio: fix multiple choice answers

jdoutreloux:
The dictionnary with the answer for a multiple choice question should only contain true answers.


benwillig (approved):
LGTM

PR #104, opened 2023-11-29
[15.0][FIX] survey_resource_booking: adapt to partner refactor

chienandalu:
We need to adapt tests to https://github.com/OCA/calendar/pull/110

cc @Tecnativa TT45338

please review @carolinafernandez-tecnativa @pedrobaeza

### DIFF
--- a/survey_resource_booking/tests/test_invite.py
+++ b/survey_resource_booking/tests/test_invite.py
@@ -32,7 +32,7 @@ class SurveyInvitationCase(TransactionCase):
         # Make sure RB gets surveyed as expected
         booking_f = Form(self.env["resource.booking"])
         booking_f.type_id = self.rbt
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
Fast-tracking this as obvious:

/ocabot merge patch

chienandalu:
Also failing affecting to the tests in sale_resource_booking and website_sale_resource_booking (investigating...)

pedrobaeza:
One way to avoid this compatibility problem is to add an inverse function to `partner_id` field, so that it put on `partner_ids` the proper value. Do you know what I mean?

chienandalu:
Yes, I was thinking about that, although in this case the problem is that using the `Form` class, the field should be in the view

pedrobaeza:
Not sure if putting it as invisible will serve.

PR #105, opened 2024-01-10
[15.0][ADD] survey_partner_representative: New module

chienandalu:
When we have the modules `survey_contact_generation`, `survey_sale_generation` or `survey_crm_generation` a salesman or a portal users could want to act as a representative of another using the surveys that generate those documents as an entry point.

Pending:

- [x] Tests

cc @Tecnativa TT46965

ready for review @pedrobaeza @victoralmau


pedrobaeza:
It seems a temporary deb file has slipped into the PR.

chienandalu:
Oh, weird, maybe it has to do with the new fragments conversion

survey_partner_representative/models/survey_survey.py
```diff
@@ -0,0 +1,20 @@
...
+        help="A user with the proper permissions, could do the survey on behalf of other"
+    )
+
+    def _create_answer(self, *args, **kwargs):
+        """Inject the context so the partner representation is only done on a normal
+        survey workflow"""
+        if self.allow_partner_representing:
+            return super(
+                SurveySurvey, self.with_context(survey_partner_representative=True)
+            )._create_answer(*args, **kwargs)
```
pedrobaeza: This way is more compact and readable:

```suggestion
        if self.allow_partner_representing:
            self = self.with_context(survey_partner_representative=True)
```

survey_partner_representative/models/survey_representative_mixin.py
```diff
@@ -0,0 +1,27 @@
...
+    survey_representative_partner_id = fields.Many2one(
+        comodel_name="res.partner",
+        compute="_compute_survey_representative_partner_id",
+        store=True,
+    )
+
+    @api.depends("survey_user_input_id")
+    def _compute_survey_representative_partner_id(self):
+        for record in self:
+            record.survey_representative_partner_id = (
```
pedrobaeza, line 27: I suppose this one can't be a related due to the initialization, so put a comment about that for not being tempted in the future to change it. Not being covered in tests may mean dead code?

pedrobaeza (approved):

survey_partner_representative/models/survey_representative_mixin.py
chienandalu: The idea is that the field can be user as a regular one (users can set it) and this mixin is meant for glue modules that would need this same behavior.

For testing part, I gues it could be done somehow, I should look for a runtime model patching or something like that...

pedrobaeza: OK, got it.

pedrobaeza: `odoo_test_helper` library is for that things, but it's not worth right now I think.

chienandalu: Oh, I just added a little test for that coverage :)

survey_partner_representative/__manifest__.py
```diff
@@ -0,0 +1,28 @@
...
+    "license": "AGPL-3",
+    "depends": ["survey"],
+    "data": [
+        "security/survey_partner_representative_security.xml",
+        "views/survey_survey_views.xml",
+        "views/survey_user_input_views.xml",
+    ],
+    "demo": ["demo/survey_partner_representative_demo.xml"],
+    "external_dependencies": {
+        "python": ["odoo_test_helper"],
```
pedrobaeza: Don't add it here. It's only for tests. Instead, create a file `test_requirements.txt` for being installed on runboat and CI.

pedrobaeza (approved):

victoralmau (approved):

PR #108, opened 2024-01-19
[15.0][IMP] survey_skip_start: simpler implementation

chienandalu:
Simplifies the implementation and bring compatibility with any survey layout.

cc @Tecnativa TT47278


survey_skip_start/models/survey_survey.py
```diff
@@ -12,13 +11,7 @@ class SurveySurvey(models.Model):
...
-        """For the moment, skipping the start screen is only compatible with single
-        page layout surveys"""
-        if self.filtered(lambda x: x.skip_start and x.questions_layout != "one_page"):
-            raise UserError(
-                _(
-                    "The skip start screen option is only compatible with on page layouts"
-                )
-            )
+    def _create_answer(self, *args, **additional_vals):
+        inputs = super()._create_answer(*args, **additional_vals)
```
pedrobaeza, line 17: Please docstring about why are you overwriting this method.

pedrobaeza (commented):
As this is hacking current process, you should document it properly for future references, and make sure the test covers the flow in case this fails.

survey_skip_start/models/survey_survey.py
chienandalu: done :+1:

pedrobaeza (approved):
/ocabot merge patch

PR #110, opened 2024-02-08
[15.0][IMP] survey_contact_generation: parent company

chienandalu:
We can use the new survey setting *Create parent contact* so when we have a question linked to the `company_name` field we can create a parent to the generated contact.

cc @Tecnativa TT47577

please review @pilarvargas-tecnativa @pedrobaeza


pedrobaeza (approved):

pilarvargas-tecnativa (approved):
:+1:

PR #113, opened 2024-02-13
[15.0][IMP] survey_contact_generation: link partner

chienandalu:
For some advanced workflows is quite convenient to have the linked input on the generated contacts.

cc @Tecnativa TT47573

### DIFF
--- a/survey_contact_generation/models/res_partner.py
+++ b/survey_contact_generation/models/res_partner.py
@@ -1,0 +1,9 @@
+# Copyright 2024 Tecnativa - David Vidal
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    generating_survey_user_input_id = fields.Many2one(comodel_name="survey.user_input")
--- a/survey_contact_generation/models/__init__.py
+++ b/survey_contact_generation/models/__init__.py
@@ -1,3 +1,4 @@
+from . import res_partner
 from . import survey_question
 from . import survey_survey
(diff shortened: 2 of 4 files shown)


survey_contact_generation/models/res_partner.py
```diff
@@ -0,0 +1,9 @@
+# Copyright 2024 Tecnativa - David Vidal
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    survey_user_input_id = fields.Many2one(comodel_name="survey.user_input")
```
pedrobaeza: Shouldn't be the reverse direction? Put the generated partner in the survey_input?

chienandalu: Already is, but I wan't also to flag those generated partners so I can easily discriminate them for other workflows (look at the tt task for an example)

pedrobaeza: And what about the inverse one2many?

chienandalu: I just want to link inputs to partners generated by that input, but the input could have a partner which isn't generated (that's the normal behavior)

pedrobaeza: OK, let's try to make it more self-discovering by name:

```suggestion
    generating_survey_user_input_id = fields.Many2one(comodel_name="survey.user_input")
```

chienandalu: Done :+1:

pedrobaeza (approved):

victoralmau (approved):

PR #114, opened 2024-02-15
[fix] survey_question_type_five_star: remove reference to style from other module (nps)

ni-ka:
removed reference to $nps-widget-color; which caused an error whenever the survey is viewed (survey still worked). This fix removes the error


chienandalu (changes requested):
Check pre-commit errors

PR #92, opened 2023-08-18
[15.0][ADD] survey_answer_generation: New module

chienandalu:
cc @Tecnativa TT40984

please review @pedrobaeza @pilarvargas-tecnativa

PR #115, opened 2024-02-26
[15.0][ADD] survey_next_survey_update_partner: new module

chienandalu:
This module acts as a bridge of `survey_contact_generation` and `survey_answer_generation` so
when we link contact generation questions, the values that change in the next survey update
those of the created partner.

pending: 
- [x] https://github.com/OCA/survey/pull/92

cc @Tecnativa TT47575


arielbarreiros96:
@chienandalu I was reviewing this workflow, and while it is quite interesting that the answers are pre-filled from the beginning. Wouldn't it be interesting to add this update feature directly in `survey_contact_generation` and keep this two features separate as they are now?
At the moment this module (`survey_contact_generation `) has the `generate_contact` field, there could be for instance an `update_contact` field that checks if it will be possible to update or not, and if there is a contact with the same email then update it.
In [this ](https://github.com/OCA/survey/blob/a02bfe54d7ad0c2772c38cd6c664c409d305e69d/survey_contact_generation/models/survey_user_input.py#L79) method you already do all the checks and create the contact, from here it should not be too complicated to update the existing partner.
I would appreciate your thoughts on this, does it make sense to you? Thanks in advance

chienandalu:
Hi @arielbarreiros96 the thing is that updating the contact makes no sense in a regular workflow without linked surveys. Just here, where we have a subsequent survey which answers are linked to the original one that generated a brand new contact...

Not the cleanest workflow... but it's our customer's requirements... :man_shrugging:

pilarvargas-tecnativa (approved):

pedrobaeza (approved):
/ocabot merge nobump

PR #117, opened 2024-03-08
[16.0] [MIG] survey_formio

ThomasBinsfeld:
From https://github.com/acsone/survey/tree/13.0/survey_formio


survey_formio/models/survey.py
```diff
@@ -84,25 +78,16 @@ def _from_formio_multiple_choice_to_odoo(
...
         multiple_choice_answers_dict: dict given by form.io for a multiple choice question
         """
-        answers_dict = {}
-        multiple_choice_answers_dict = {
-            key: value for (key, value) in multiple_choice_answers_dict.items() if value
-        }
-        for answer in multiple_choice_answers_dict.keys():
-            answers_dict[question_id + "_" + answer[1:]] = answer[1:]
-        return answers_dict
+        return [value for value in multiple_choice_answers_dict.values() if value]
```
jdoutreloux: I've tested this and it seems there's an issue here. 
Saving a multiple choice answer given by formio doesn't work.  I think Odoo is waiting for the id of the answer, not its value (see : https://github.com/odoo/odoo/blob/16.0/addons/survey/tests/test_survey_ui_session.py#L210 ) 
I think this should be 
`return [value for value in multiple_choice_answers_dict.keys() if value]` or something like that

ThomasBinsfeld: Should be better now

jdoutreloux: It works now. Thanks :)

jdoutreloux (approved):

survey_formio/models/survey.py
ThomasBinsfeld: Thanks for the review!

FrancoMaxime (approved):
LGTM:  code review

dreispt (approved):

PR #118, opened 2024-03-14
[15.0][IMP] survey_contact_generation: partner hook method

chienandalu:
Added a hook method so we can easily override the partner match.

cc @Tecnativa 
please check @pedrobaeza @victoralmau

### DIFF
--- a/survey_contact_generation/models/survey_user_input.py
+++ b/survey_contact_generation/models/survey_user_input.py
@@ -72,6 +72,10 @@ class SurveyUserInput(models.Model):
             subtype_id=self.env.ref("mail.mt_note").id,
         )
 
+    def _get_existing_partner(self, email):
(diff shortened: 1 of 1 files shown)


victoralmau (approved):

survey_contact_generation/models/survey_user_input.py
```diff
@@ -72,6 +72,11 @@ def _create_contact_post_process(self, partner):
             subtype_id=self.env.ref("mail.mt_note").id,
         )
 
+    @api.model
```
pedrobaeza: This shouldn't be `api.model`, as you may want to use `self` for selecting the partner.

chienandalu: You're right. Fixed

pedrobaeza (approved):
/ocabot merge patch

PR #119, opened 2024-04-15
[15.0][FIX] survey_result_mail: sending condition

chienandalu:
We need to group the mail condition so it doesn't send the mail when the last one is true.

cc @Tecnativa TT48729

please check @pedrobaeza @victoralmau


victoralmau (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #130, opened 2024-08-22
[16.0][ADD] module survey_odoo_debranding

mathben:
- remove Powered by Odoo into survey website


chienandalu (changes requested):
Check CI errors

chienandalu:
I recommend you setting up pre-commit in your local to fix it in advance to pushing: https://github.com/OCA/maintainer-tools/wiki/Install-pre-commit

survey_odoo_debranding/templates/disable_odoo.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+    <!-- Remove "Powered by Odoo" -->
+    <template
+        id="frontend_layout_remove_brand"
+        inherit_id="survey.layout"
```
mathben, line 6: I remove customize_show=True
I need to retest it

mathben: Test done, it works without customize_show=True

mathben:
Thanks for your support @chienandalu 
About the test cover, I am base with this module : https://github.com/OCA/website/blob/16.0/website_odoo_debranding/hooks.py#L16
What do you think?
And this sentence "- XML changes in website are made using 'customize_show=True'" is not good anymore, because I remove customize_show=True because it fails "test / test with OCB (pull_request)"

chienandalu:
I'd say the issue was that `customize_show` is a field introduced by `portal` and `survey` doesn't have it in its dependency chain. In any case, you don't need for the task.

For the test cover, it isn't blocking.

chienandalu (approved):
Tested :+1:

mathben:
Thanks.
So, I remove this comment into hooks.py
`# - XML changes in website are made using 'customize_show=True'`

PR #146, opened 2025-02-11
[15.0][ADD] survey_certification_sending: New module

pilarvargas-tecnativa:
cc @Tecnativa TT54988

@chienandalu @carlos-lopez-tecnativa please review


survey_certification_sending/models/survey_survey.py
```diff
@@ -0,0 +1,12 @@
+# Copyright 2025 Tecnativa - Pilar Vargas
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import fields, models
+
+
+class Survey(models.Model):
+    _inherit = "survey.survey"
+
+    skip_auto_email = fields.Boolean(
```
chienandalu:
```suggestion
    skip_certification_email = fields.Boolean(
```

chienandalu (commented):
Very quick review for the moment

survey_certification_sending/models/survey_survey.py
pilarvargas-tecnativa: thanks

survey_certification_sending/models/survey_user_input.py
```diff
@@ -0,0 +1,55 @@
...
+                    user_input.certification_sent = True
+        # Mark as completed and send email only if not deactivated.
+        return super(
+            SurveyUserInput,
+            self.with_context(
+                skip_certification_email=self.mapped("survey_id").filtered(
+                    lambda s: s.skip_certification_email
+                )
+            ),
+        )._mark_done()
```
chienandalu: You shouldn't avoid the survey completion, just the mail sending:


https://github.com/OCA/OCB/blob/eab555dc127b3a60a0f2b019c39abfbb3b078ff5/addons/survey/models/survey_user_input.py#L219

survey_certification_sending/models/survey_user_input.py
```diff
@@ -0,0 +1,55 @@
...
+            if (
+                not user_input.survey_id.skip_certification_email
+                and user_input.survey_id.certification
+                and user_input.scoring_success
+            ):
+                if (
+                    user_input.survey_id.certification_mail_template_id
+                    and not user_input.test_entry
+                ):
+                    user_input.certification_sent = True
```
chienandalu: This is to be flagged just when the certification mail is sent

chienandalu (changes requested):
Also we should avoid the download certification button. The purpose is to let the certification reviewer check the answers before sending letting the user get the certification

image

survey_certification_sending/models/survey_user_input.py
pilarvargas-tecnativa: It has been discussed in an internal conversation to reach the agreement to indicate with a comment this operation at this point in the logic.

survey_certification_sending/models/survey_user_input.py
pilarvargas-tecnativa: I have made the change to inject the correct ids into the context instead of the flag since it can lead to error if at some point a set of records with different values for that flag is treated.

pilarvargas-tecnativa:
Button conditionally hidden

image

survey_certification_sending/views/survey_user_views.xml
```diff
@@ -0,0 +1,40 @@
...
+        <field name="model_id" ref="survey.model_survey_user_input" />
+        <field name="state">code</field>
+        <field name="code">
+            if records:
+                result = records.action_manual_send_certification()
+                action = {
+                    "type": "ir.actions.client",
+                    "tag": "display_notification",
+                    "params": {
+                        "title": "Certification Sent",
```
carlos-lopez-tecnativa: Could this text be translated?
Otherwise, I think it could be an option for the `action_manual_send_certification` function to return the action and handle the translation within the function.

chienandalu: Agree

pilarvargas-tecnativa: Thanks!

carlos-lopez-tecnativa (approved):

victoralmau (approved):
Code review OK

PR #144, opened 2025-01-16
[15.0][MIG] survey_question_type_binary

Tardo:
Backport from 16.0

PR #147, opened 2025-02-24
[15.0][ADD] survey_question_type_binary_result_mail

Tardo:
This module adds compatibility between the survey_question_type_binary and suervey_result_mail modules.

I don't know if it would be a good idea to give the option to not send files as email attachments.
With this module, the names of the files are displayed as an answer to the “upload a file” questions and the files are attached to the mail.

depends on: https://github.com/OCA/survey/pull/144


pedrobaeza (approved):
/ocabot merge nobump

PR #156, opened 2025-03-27
[15.0][ADD] survey_certification_py3o: New module

pilarvargas-tecnativa:
cc @Tecnativa TT55652


survey_certification_py3o/models/survey_user_input.py
```diff
@@ -0,0 +1,11 @@
+# Copyright 2025 Tecnativa - Pilar Vargas
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import fields, models
+
+
+class SurveySurvey(models.Model):
+    _inherit = "survey.user_input"
+
+    py3o_template_id = fields.Many2one(
+        related="survey_id.py3o_template_id", store=True, readonly=True
```
chienandalu: Do we need it to be stored?

chienandalu (commented):
Looking good :D 

Small comment:

survey_certification_py3o/models/survey_user_input.py
pilarvargas-tecnativa: Yes
https://github.com/OCA/reporting-engine/blob/748884f88144611b51870105e9dc6a32d1b01a26/report_py3o/models/py3o_report.py#L169

chienandalu: Are you sure? I did a quick test with a related unstored field and it showed up in its model `_fields`

pilarvargas-tecnativa: right, change done!

victoralmau:
Can you check the errors in the tests?

pedrobaeza (approved):
/ocabot merge nobump

PR #160, opened 2025-04-09
[17.0][MIG] survey_contact_generation: Migration to version 17.0

pilarvargas-tecnativa:
cc @Tecnativa TT51947

@carlos-lopez-tecnativa @juancarlosonate-tecnativa please review


juancarlosonate-tecnativa:
when i do a live session, this error appears 
image
And if you try to copy to the clipboard, another error shows up and it goes into recurrence, very strange

pilarvargas-tecnativa:
> when i do a live session, this error appears 

It must be a js error in the runbot, locally it doesn't reproduce that error when creating a live session. Anyway to test the survey just click ‘test’ or share and copy the code in an incognito window without user identified.

pedrobaeza:
I think it's because of HTTP. Switch in the URL http > https to try.

survey_contact_generation/models/res_partner.py
```diff
@@ -0,0 +1,9 @@
+# Copyright 2024 Tecnativa - David Vidal
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    generating_survey_user_input_id = fields.Many2one(comodel_name="survey.user_input")
```
carlos-lopez-tecnativa:
```suggestion
    generating_survey_user_input_id = fields.Many2one(comodel_name="survey.user_input", copy=False, readonly=True)
```

survey_contact_generation/views/survey_question_views.xml
```diff
@@ -0,0 +1,32 @@
...
+            <xpath expr="//field[@name='suggested_answer_ids']" position="attributes">
+                <attribute
+                    name="context"
+                >{'default_question_id': id, 'default_res_partner_field': res_partner_field}</attribute>
+            </xpath>
+            <xpath
+                expr="//field[@name='suggested_answer_ids']//field[@name='value']"
+                position="after"
+            >
+                <field name="res_partner_field" invisible="1" />
```
carlos-lopez-tecnativa:
```suggestion
                <field name="res_partner_field" column_invisible="1" />
```

image

survey_contact_generation/views/survey_question_views.xml
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="survey_question_form" model="ir.ui.view">
+        <field name="model">survey.question</field>
+        <field name="inherit_id" ref="survey.survey_question_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='comments_allowed']/.." position="after">
+                <group name="contact" string="Contact">
+                    <field name="res_partner_field" widget="selection" />
```
carlos-lopez-tecnativa: Why use the `selection` widget?

The widget can be removed, and options can be added to achieve similar behavior while allowing the user to search within the field. What do you think?

```suggestion
                    <field name="res_partner_field" options="{'no_create': True, 'no_open': True}" />
```

survey_contact_generation/models/survey_question.py
```diff
@@ -0,0 +1,79 @@
...
+            record.allowed_field_ids = (
+                self.env["ir.model.fields"]
+                .search(
+                    [
+                        ("model", "=", "res.partner"),
+                        ("ttype", "in", type_mapping.get(record.question_type, [])),
+                    ]
+                )
+                .ids
+            )
```
carlos-lopez-tecnativa: I think this code can be improved. Now it's possible to add a domain based on another field without adding a many2many field. Could you check if the example below is applicable here?

https://github.com/odoo/odoo/blob/01566655befd795071fdc6a3c4d22605e1c8b262/addons/account/models/account_tax.py#L1738

https://github.com/odoo/odoo/blob/01566655befd795071fdc6a3c4d22605e1c8b262/addons/account/views/account_tax_views.xml#L47

survey_contact_generation/models/survey_user_input.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def _get_existing_partner(self, email, limit=1):
+        """Hook method that can be used to change the behavior of contact generation"""
+        return self.env["res.partner"].search([("email", "=ilike", email)], limit=limit)
+
+    def _mark_done(self):
+        """Generate the contact when the survey is submitted"""
+        for user_input in self.filtered(
+            lambda r: r.survey_id.generate_contact and not self.partner_id
+        ):
```
carlos-lopez-tecnativa: Sorry, but this is not working for me. I completed the survey, but the contact was not created. I'm not sure if the filter is correct. Please check my suggestion or investigate why this isn't working, or let me know if I missed any step or configuration.

```suggestion
        for user_input in self.filtered(
            lambda r: r.survey_id.generate_contact and not r.partner_id
        ):
```

image
image
image
image

carlos-lopez-tecnativa (changes requested):

survey_contact_generation/models/survey_question.py
pilarvargas-tecnativa: I have adapted the code to your suggestion and it seems to work correctly. Could you check again?

survey_contact_generation/models/survey_question.py
```diff
@@ -0,0 +1,73 @@
...
+            "date": ["date", "text", "char"],
+            "datetime": ["datetime", "html", "char"],
+            "simple_choice": ["many2one", "html", "char"],
+            "multiple_choice": ["many2many", "html", "char"],
+        }
+        for record in self:
+            allowed_types = type_mapping.get(record.question_type, [])
+            domain = [
+                ("model", "=", "res.partner"),
+                ("ttype", "in", allowed_types),
```
carlos-lopez-tecnativa, line 32: I  think it's necessary to add a domain to only permit selected `stored` fields (and discard computed or related ones with `store=False`). What do you think?

image

carlos-lopez-tecnativa (changes requested):
Sorry, but in this case I can't test it on Runboat, because even though the survey is configured to not require login, when I copy the link, it redirects to the login page. I don't understand why. Could you review it, please?
image


image

survey_contact_generation/models/survey_question.py
pilarvargas-tecnativa: Added, thanks

pilarvargas-tecnativa:
> Sorry, but in this case I can't test it on Runboat, because even though the survey is configured to not require login, when I copy the link, it redirects to the login page. I don't understand why. Could you review it, please? 

There seems to be a problem with runbot and it won't let you test in an unknown window. 

In local there is no problem, in runbot my steps have been to copy the link of the "Share" button of the survey:

http://oca-survey-17-0-pr160-0e7573dd82f9.runboat.odoo-community.org/survey/start/80e5f1e2-1a9d-4c51-8e23-09e93f7fa2c5

Logout and paste the link in the browser without being in an unknown session.

image
image

carlos-lopez-tecnativa (approved):
LGTM

juancarlosonate-tecnativa (approved):
LGTM! As you said, if you copy the share link, log out, and then do the survey, it creates the contact correctly. Thx!

PR #168, opened 2025-05-07
[16.0][IMP] partner_survey: add a wizard to share surveys with partners

kobros-tech:


pedrobaeza (commented):
Why not doing this on the list  for sharing the survey with several partners?

kobros-tech:
> Why not doing this on the list for sharing the survey with several partners?

good idea, we can extend!

partner_survey/views/res_partner_view.xml
```diff
@@ -57,6 +57,18 @@
                     </div>
                 </button>
             </xpath>
+
```
pedrobaeza: Don't add empty lines inside.

partner_survey/wizard/take_survey.py
```diff
@@ -0,0 +1,28 @@
+from odoo import _, fields, models
+
+
+class take_survey_wizard(models.Model):
+    """Wizard allowing to take a survey from partners form view"""
+
+    _name = "res.partner.survey.survey.wizard"
+
+    partner_ids = fields.Many2many("res.partner")
```
pedrobaeza: Should this be editable?

```suggestion
    partner_ids = fields.Many2many("res.partner", required=True)
```

partner_survey/wizard/take_survey.xml
```diff
@@ -0,0 +1,56 @@
...
+    </record>
+
+    <record id="action_res_partner_survey_survey_wizard" model="ir.actions.act_window">
+        <field name="name">Share a Survey Current Partner</field>
+        <field name="res_model">res.partner.survey.survey.wizard</field>
+        <field name="view_id" ref="partner_survey.res_partner_survey_survey_wizard" />
+        <field name="target">new</field>
+        <field name="context">{
+            'default_partner_ids': active_ids
+        }</field>
```
pedrobaeza:
```suggestion
        <field name="context">{'default_partner_ids': active_ids}</field>
```

partner_survey/wizard/take_survey.xml
```diff
@@ -0,0 +1,56 @@
...
+        id="action_res_partner_survey_survey_multilist_wizard"
+        model="ir.actions.act_window"
+    >
+        <field name="name">Share a Survey with Selected Partners</field>
+        <field name="res_model">res.partner.survey.survey.wizard</field>
+        <field name="view_mode">form</field>
+        <field name="target">new</field>
+        <field name="context">{
+            'default_partner_ids': active_ids
+        }</field>
```
pedrobaeza:
```suggestion
        <field name="context">{'default_partner_ids': active_ids}</field>
```

partner_survey/wizard/take_survey.xml
```diff
@@ -0,0 +1,56 @@
...
+                        special="cancel"
+                        data-hotkey="z"
+                        class="btn-secondary"
+                    />
+                </footer>
+            </form>
+        </field>
+    </record>
+
+    <record id="action_res_partner_survey_survey_wizard" model="ir.actions.act_window">
```
pedrobaeza, line 31: Why do you need 2 action windows?

partner_survey/wizard/take_survey.py
pedrobaeza: What about this?

pedrobaeza (approved):

wlin-kencove:
image


@kobros-tech 

When I try the button on runboat, I see the wizard when I press the button on partner but there are no surveys to select. 
There is a blank one but I cannot continue with the form afterwords. 

See image above.

kobros-tech:
@wlin-kencove 

hi WL, make sure to have some surveys created, at least one or two to select from.

wlin-kencove:
@kobros-tech 

image


image


See above from runboat. There are surveys defined by default but I added a few to make sure. When I  try to share from customer form it does not display any. 

Same behavior occurs when I try to share via listview:

image

wlin-kencove:
@kobros-tech Perhaps it is permissions related if you are not seeing the same issues? 

image


User should be admin on runboat.

kobros-tech:
> @kobros-tech Perhaps it is permissions related if you are not seeing the same issues?
> 
> image
> User should be admin on runboat.

all right, the general base user is accessing and editing the wizard, let us see if reviewers have the same issue or any other.

wlin-kencove:
@kobros-tech 

Same behavior when I tested locally

image

kobros-tech:
@wlin-kencove 

this issue had one solution which is add view ref xml id, it should not be necessary but it seems that no rule is dominant.

wlin-kencove:
@kobros-tech  Looks good now. Thanks!

wlin-kencove (approved):

PR #172, opened 2025-05-23
[15.0][FIX] survey_certification_py3o: Change conditions to generate report + test

pilarvargas-tecnativa:
Checking only the model can generate an error when there is no template defined, better to check if the template is defined too. The model that generates the report is actually survey.user_input

@tecnativa 

@victoralmau @carlos-lopez-tecnativa please review

### DIFF
--- a/survey_certification_py3o/models/py3o_report.py
+++ b/survey_certification_py3o/models/py3o_report.py
@@ -9,6 +9,9 @@ class Py3oReport(models.TransientModel):
     _inherit = "py3o.report"
 
     def _get_template_fallback(self, model_instance):
-        if model_instance._name == "survey.survey":
+        if (
+            model_instance._name == "survey.user_input"
+            and model_instance.py3o_template_id
+        ):
             return b64decode(model_instance.py3o_template_id.py3o_template_data)
         return super()._get_template_fallback(model_instance)
--- a/survey_certification_py3o/tests/test_survey_certification_py3o.py
+++ b/survey_certification_py3o/tests/test_survey_certification_py3o.py
@@ -1,8 +1,33 @@
+import io
+import os
+import zipfile
+from base64 import b64encode
+
+from odoo.modules.module import get_module_resource
+
 from odoo.addons.survey.tests.common import TestSurveyCommon
 
 
 class TestCertificationPy3o(TestSurveyCommon):
     def test_certification_py3o(self):
+        demo_odt_path = get_module_resource(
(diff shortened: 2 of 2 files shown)


survey_certification_py3o/models/py3o_report.py
```diff
@@ -9,6 +9,9 @@ class Py3oReport(models.TransientModel):
     _inherit = "py3o.report"
 
     def _get_template_fallback(self, model_instance):
-        if model_instance._name == "survey.survey":
+        if (
```
pedrobaeza, line 12: This is too generic and can trigger undesired results. I still prefer the `_name` condition. If the model is `survey.user_input`, then change it, but I saw the other model and it's logical to define the certification at survey level, not answer level.

pedrobaeza (changes requested):

survey_certification_py3o/models/py3o_report.py
pilarvargas-tecnativa: OK, I add the model, although the report action is only for survey.user_input. If we don't check that the field exists, the error if someone tries to print that report manually is not controlled.
image

However, if the field is checked, the error is controlled by the module itself.
https://github.com/OCA/reporting-engine/blob/15.0/report_py3o/models/py3o_report.py#L173
image

The template is defined in the survey.survey, but the report is from the certification so in the suvey_user_input model we have a related to the survey template.

pedrobaeza: But if this error happens, then we are not putting the "related" in the proper module, which is here. Please do that.

pilarvargas-tecnativa: The related is correct. The error arises when you try to manually print a report, normally the report is sent by mail and then you check that there is a custom template for the survey and then it will be prioritised in the ‘automatic’ part. But there is also the possibility of printing the report manually from a report action, that is where you have to control this error because there may be surveys that do not have that custom template defined. But that's all for the manual part, not the odoo flow when getting a certification (email and download button in frontend).

image

image

In conclusion, this module provides the possibility to have a customised template for printing certifications at survey level. If this template exists in the survey, it will automatically be the one that is sent by mail (instead of the odoo one) and the one that the user downloads after finishing the survey. In addition, the possibility of printing this report manually from the answers of the certification is added (this is where you have to control the error).

pedrobaeza: OK, understood, thanks.

pedrobaeza (approved):
/ocabot merge patch

PR #173, opened 2025-05-23
[17.0][MIG] survey_certification_py3o: Migration to version 17.0

pilarvargas-tecnativa:
@Tecnativa  TT56326

@carlos-lopez-tecnativa @victoralmau please review


carlos-lopez-tecnativa (changes requested):
When a user tries to print a report but the survey doesn't have a template, I get this error.

Could we handle the error and raise a `UserError` instead of showing a traceback?image

Additionally, please squash the administrative commits.

pilarvargas-tecnativa:
> When a user tries to print a report but the survey doesn't have a template, I get this error.

This is a error handled in the report_py3o module. The priority of this module is to send the custom report in case of a defined template. Manual printing offers the possibility of downloading it from the backend by someone in charge to provide a copy or send it outside of odoo, although this should not be the usual flow.

carlos-lopez-tecnativa (approved):
> This is a error handled in the report_py3o module. The priority of this module is to send the custom report in case of a defined template. Manual printing offers the possibility of downloading it from the backend by someone in charge to provide a copy or send it outside of odoo, although this should not be the usual flow.

OK, I understand. Although I’d prefer to get a more user-friendly error instead of a traceback, but anyway, let’s go.

survey_certification_py3o/controllers/main.py
```diff
@@ -11,10 +10,19 @@ def _generate_report(self, user_input, download=True):
...
-            ._render_py3o([user_input.id], data={"report_type": "pdf"})[0]
+            request.env["ir.actions.report"]
+            .sudo()
+            ._render_py3o(
+                "survey_certification_py3o.custom_certification_report",
+                [user_input.id],
+                data={"report_type": "pdf"},
+            )[0]
         )
+        # report = (
```
pedrobaeza: Please remove this commented code.

survey_certification_py3o/models/py3o_report.py
```diff
@@ -9,6 +9,9 @@ class Py3oReport(models.TransientModel):
     _inherit = "py3o.report"
 
     def _get_template_fallback(self, model_instance):
-        if model_instance._name == "survey.survey":
+        if (
+            model_instance._name == "survey.user_input"
+            and model_instance.py3o_template_id
```
pedrobaeza, line 14: Wasn't this `model_instance.survey_id.py3o_template_id`?

pilarvargas-tecnativa: https://github.com/OCA/survey/blob/15.0/survey_certification_py3o/models/survey_user_input.py#L9

The template is set at survey level but the related of survey.user_input is checked.

v15: https://github.com/OCA/survey/blob/15.0/survey_certification_py3o/models/py3o_report.py#L12-L15

victoralmau (approved):
Code review OK

pedrobaeza (approved):
/ocabot merge nobump

PR #180, opened 2025-06-16
[16.0][ADD] survey_start_end_text

kobros-tech:


survey_start_end_text/views/survey_templates.xml
```diff
@@ -0,0 +1,24 @@
...
+            <attribute name="t-if">False</attribute>
+        </xpath>
+        <xpath expr="//button//t[@t-if='False']" position="after">
+            <t t-if="survey.start_button_name">
+                <t t-out="survey.start_button_name" />
+            </t>
+            <t t-elif="survey.certification">
+                Start Certification
+            </t>
+        </xpath>
```
adasatorres, line 14: Good morning, is it necessary to set the if condition to False in order to add it again later? 

I suggest the following change:

```suggestion
        <xpath expr="//button/t[@t-if='survey.certification']" position="after">
           <t t-elif="survey.start_button_name">
               <t t-out="survey.start_button_name" />
            </t>
        </xpath>
```
This should also account for the "certification" option.

adasatorres (changes requested):

PR #181, opened 2025-06-19
[17.0][MIG] survey_resource_booking: Migration to 17.0

victoralmau:
Migration to 17.0

Please @pedrobaeza and @carlos-lopez-tecnativa can you review it?

@Tecnativa TT51955


pedrobaeza (approved):

survey_resource_booking/models/resource_booking.py
```diff
@@ -0,0 +1,61 @@
...
+
+class ResourceBooking(models.Model):
+    _inherit = "resource.booking"
+
+    survey_user_input_id = fields.Many2one(
+        comodel_name="survey.user_input",
+        string="Survey user input",
+        tracking=True,
+        help="User responses to survey, as specified in the resource booking type.",
+    )
```
carlos-lopez-tecnativa: I think this field should not be editable by the user, or at the very least, it should not be copied. What do you think?
image

```suggestion
    survey_user_input_id = fields.Many2one(
        comodel_name="survey.user_input",
        string="Survey user input",
        tracking=True,
        copy=False,
        readonly=True,
        help="User responses to survey, as specified in the resource booking type.",
    )
```

carlos-lopez-tecnativa (approved):
LGTM, just a minor change if you consider it applicable (non-blocking comment).

survey_resource_booking/models/resource_booking.py
victoralmau: `copy=false` added.

Although I understand what you indicate that this field is readonly in the view, perhaps in some cases it may be interesting to change it, therefore, I think it is better to leave it as it is in case someone is using this possibility.

PR #184, opened 2025-07-03
[16.0][ADD] survey_signature

adasatorres:
Good morning, this addon creates a signature-type question for the survey module. It is configured like any other type of field.


sheragar (approved):
LGTM. Funtional Review 👍

rrebollo (commented):
I think `setup` folder shouldn't be part of PR

Image

adasatorres:
Hi @rrebollo, if I recall correctly, those setup modifications are generated by one of the pre-commit hooks. It’s normal for these files to be automatically created in the setup directory when the addon is new in the repository, at least in version 16. In later versions, this behavior might have changed.

In any case, perhaps someone else has more knowledge about this and can help clarify it for us.

rrebollo:
@adasatorres sorry for making you waste your time. I'm on my phone now but looking at already merged code It's is more like You are completely right.

survey_signature/models/survey_question.py
```diff
@@ -0,0 +1,9 @@
+# Copyright 2025 Binhex - Zuzanna Elżbieta Szalaty Szalaty
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from odoo import _, fields, models
+
+
+class SurveyQuestion(models.Model):
+    _inherit = "survey.question"
+
+    question_type = fields.Selection(selection_add=[("signature", _("Signature"))])
```
rrebollo, line 9: What about setting a fallback value here?

survey_signature/models/survey_user_input.py
```diff
@@ -0,0 +1,23 @@
...
+
+class SurveyUserInput(models.Model):
+    _inherit = "survey.user_input"
+
+    def save_lines(self, question, answer, comment=None):
+        old_answers = self.env["survey.user_input.line"].search(
+            [("user_input_id", "=", self.id), ("question_id", "=", question.id)]
+        )
+        if question.question_type in ["signature"]:
+            self._save_line_signature(question, old_answers, answer)
```
rrebollo, line 15: This path never calls super. Not my favorite approach but I don't know enough about the built-in survey, I'm guessing there isn't another way.

survey_signature/models/survey_user_input.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2025 Binhex - Zuzanna Elżbieta Szalaty Szalaty
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from odoo import models
+
+
+class SurveyUserInput(models.Model):
+    _inherit = "survey.user_input"
+
+    def save_lines(self, question, answer, comment=None):
+        old_answers = self.env["survey.user_input.line"].search(
```
rrebollo, line 11: Insert ‘self.ensure_one()’ before ???

survey_signature/models/survey_user_input_line.py
```diff
@@ -0,0 +1,20 @@
...
+# Copyright 2025 Binhex - Adasat Torres de León
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from odoo import _, api, fields, models
+
+
+class SurveyUserInputLine(models.Model):
+    _inherit = "survey.user_input.line"
+
+    value_signature = fields.Binary("Signature")
+    answer_type = fields.Selection(selection_add=[("signature", _("Signature"))])
```
rrebollo, line 11: Consider setting the ondelete attribute for the fallback value

survey_signature/models/survey_user_input_line.py
```diff
@@ -0,0 +1,20 @@
...
+
+    @api.depends("answer_type")
+    def _compute_display_name(self):
+        for line in self:
+            if line.answer_type == "signature":
+                line.display_name = (
+                    _("Signed") if line.value_signature else _("Not Signed")
+                )
+        return super()._compute_display_name()
```
rrebollo, line 20: 👍

rrebollo (approved):
Code Review: Great work! The code looks good to me (LGTM). Thank you for your contribution! I've provided a few suggestions for your consideration—feel free to address them as you see fit.

survey_signature/models/survey_user_input.py
adasatorres: Good morning @rrebollo, I don't understand where the problem you're telling me about is.

rrebollo: One return path in your method override doesn't propagate to `super()`, breaking the inheritance chain. That's not necessarily a bad thing — as I said, maybe it's the only viable approach. However, one branch of the `if` calls `super()`, while the other doesn't.

Did I explain myself better now?

adasatorres: Good morning @rrebollo  , in this case it's not really necessary to always call super. This method is executed once per record, and you should only call super when uploading if the question type is not "signature".

survey_signature/models/survey_question.py
adasatorres: Hi, @rrebollo, if I remember correctly, this field in the survey is not a base field, and the selection_add makes it behave as if on_delete is set to SET NULL by default, so I don't think it's appropriate to force it.

survey_signature/models/survey_user_input_line.py
adasatorres: Hi, @rrebollo, if I remember correctly, this field in the survey is not a required field, and the selection_add attribute automatically sets a default ondelete behavior, which is SET NULL. I don't see the need to force it.

edescalona (approved):
Hi @adasatorres , great work with this new type. Just check the test failure whenever you can. The same thing happens in another PR, so by fixing this one, you'll fix both.

LGTM

adasatorres:
Good morning @edescalona , the error in the tests is from another addon, it has nothing to do with this one.

PR #186, opened 2025-07-11
[16.0][FIX] survey_question_type_five_star: Add validation on frontend.

adasatorres:
Good morning, I'm leaving a small correction here, as I believe that any type of question should have its validation in the JavaScript file.


edescalona (approved):
LGTM

PR #189, opened 2025-07-22
[15.0][FIX] survey_sale_generation: use env.company instead of create_uid.ompany_id

pilarvargas-tecnativa:
When generating a quotation from a survey response, the sale order was being assigned to the company of the `create_uid`. This is incorrect in multi-company contexts, especially when the user has access to multiple companies and their default company (`company_id`) is not the active one.

Instead, this commit uses `self.env.company` to retrieve the currently active company from the context, which correctly reflects the company in which the user is operating when completing the survey.

This change ensures that the sale order is created in the company active in the user's session, avoiding multi-company integrity errors (e.g. trying to create a record in a company other than the one currently active).

image

@Tecnativa TT57057

### DIFF
--- a/survey_sale_generation/models/survey_user_input.py
+++ b/survey_sale_generation/models/survey_user_input.py
@@ -17,7 +17,7 @@ class SurveyUserInput(models.Model):
             "partner_id": self.partner_id.id or self.create_uid.partner_id.id,
             "origin": self.survey_id.title,
             "survey_user_input_id": self.id,
-            "company_id": self.create_uid.company_id.id,
+            "company_id": self.env.company.id,
             "team_id": self.survey_id.crm_team_id.id,
             "sale_order_template_id": sale_template.id,
             "user_id": (


victoralmau (approved):
It seems totally coherent.

survey_sale_generation/models/survey_user_input.py
```diff
@@ -17,7 +17,7 @@ def _prepare_quotation(self):
             "partner_id": self.partner_id.id or self.create_uid.partner_id.id,
             "origin": self.survey_id.title,
             "survey_user_input_id": self.id,
-            "company_id": self.create_uid.company_id.id,
+            "company_id": self.env.company.id,
```
pedrobaeza, line 20: Maybe you are generating the sale for one of your secondary companies. Isn't there any company data at survey or other level?

pilarvargas-tecnativa: Both the survey and the quote templates have the company_id field, but it may or may not be defined. The fact is that the quote must belong to a company, but setting the user's company is not correct, as the quote template and the survey may or may not belong to another company. For this reason, we have opted to use the company that is active at the time the quote is created.

pedrobaeza: I know current code is incorrect, but the one you propose may not be correct either. According to what you say, this should use the survey's one, and fallback to this other. 

```suggestion
            "company_id": self.survey_id.company_id.id or self.env.company.id,
```

pilarvargas-tecnativa: But then you would have to add the dependency to survey_multi_company

pedrobaeza: You haven't told me that! That's why I asked if there's the company field in any place. If there's no other company reference, then it's OK.

pilarvargas-tecnativa: That does not require dependencies, there is no other.

pedrobaeza (approved):
/ocabot merge patch

PR #191, opened 2025-07-23
[15.0][IMP-FIX] survey_multi_company: force survey company in context to avoid multi-company issues

pilarvargas-tecnativa:
In multi-company environments, survey requests made through web routes are executed without a specific company_id in the context. As a result, request.env.company defaults to the user's default company or the first in allowed_company_ids, which may lead to access errors, incorrect rule evaluations, or unexpected behavior.
To ensure that the survey logic runs under the correct company context, allowed_company_ids is explicitly set to the company assigned to the survey.

image

image

### DIFF
--- a/survey_multi_company/controllers/main.py
+++ b/survey_multi_company/controllers/main.py
@@ -1,0 +1,25 @@
+# Copyright 2025 Tecnativa - Pilar Vargas
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+
+from odoo.addons.survey.controllers.main import Survey
+
+
+class SurveyMultiCompany(Survey):
+    def _get_access_data(self, survey_token, answer_token, ensure_token=True):
+        # Web survey access in multi-company environments may lack an explicit active
+        # company, causing request.env.company to default to the user's default company
+        # or the first allowed one. This can lead to incorrect rule evaluation, access
+        # errors, or company-specific logic failures. To ensure consistent behavior, we
+        # explicitly switch to the company assigned to the survey.
+        data = super()._get_access_data(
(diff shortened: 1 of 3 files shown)


survey_multi_company/controllers/main.py
```diff
@@ -0,0 +1,31 @@
...
+        # allowed_company_ids is explicitly set to the company assigned to the survey.
+        data = super()._get_access_data(
+            survey_token, answer_token, ensure_token=ensure_token
+        )
+        survey_sudo = data.get("survey_sudo")
+        if survey_sudo and survey_sudo.company_id:
+            new_ctx = dict(request.env.context)
+            new_ctx["allowed_company_ids"] = [survey_sudo.company_id.id]
+            env = request.env(context=new_ctx)
+            data["survey_sudo"] = env["survey.survey"].browse(survey_sudo.id)
```
pedrobaeza: This is losing the `sudo`, and why rebrowse it instead of doing `with_company`?

survey_multi_company/controllers/main.py
```diff
@@ -0,0 +1,31 @@
...
+            if data.get("answer_sudo"):
+                data["answer_sudo"] = env["survey.user_input"].browse(
```
pedrobaeza: The same.

pedrobaeza (changes requested):

survey_multi_company/controllers/main.py
pilarvargas-tecnativa: thanks!

pedrobaeza (approved):

victoralmau (approved):
It seems coherent.

PR #206, opened 2025-10-17
[17.0][FIX] survey_skip_start: Cannot skip start in session survey

pilarvargas-tecnativa:
When a survey in which it has been established that the beginning of the survey should be skipped is created, a pagination error occurs because in sessions this start/waiting page cannot be skipped at the beginning of the session.

image

@Tecnativa TT58491

@chienandalu @pedrobaeza @victoralmau please review


pedrobaeza (approved):
Please change maintainer

victoralmau (approved):

PR #209, opened 2025-12-16
[17.0][ADD] survey_question_type_model_selection: new module

eduezerouali-tecnativa:
This module extends the Survey question types by introducing a new question type that allows selecting records from any model, with configurable domain filters. This makes it possible to display dynamic, filtered model-based selections directly within surveys.

@Tecnativa TT59700

ping @pilarvargas-tecnativa @carlos-lopez-tecnativa @pedrobaeza


pilarvargas-tecnativa:
It's perfect, except that even though the answers are correctly stored in the participation section, when you go to the review in the survey itself, you can't see them, only the questions.

image

For this, I want to remind you that you must add them to the view that is printed because, if I'm not mistaken, it checks the types to style them.

eduezerouali-tecnativa:
You are right. Now is displaying correctly. Thank you for your review.
Selección_046

pedrobaeza:
ping @pilarvargas-tecnativa

survey_question_type_model_selection/tests/test_question_type_drop.py
```diff
@@ -0,0 +1,39 @@
...
+                "access_mode": "public",
+                "access_token": "b137640d-14d4-4748-9ef6-344caaaaaaf",
+            }
+        )
+        cls.question = cls.env["survey.question"].create(
+            {
+                "survey_id": cls.survey.id,
+                "title": "State",
+                "question_type": "model",
+                "question_model_id": 81,
```
carlos-lopez-tecnativa: What does ID 81 correspond to? I think you should use the [_get](https://github.com/odoo/odoo/blob/11d142750637052f9fddd18e049d436078fb2bdf/odoo/addons/base/models/ir_model.py#L305) function instead of hardcoding the ID here,
like the default value `self.env["ir.model"]._get_id("model.name").`

survey_question_type_model_selection/views/survey_template.xml
```diff
@@ -0,0 +1,67 @@
...
+         <template id="survey_page_print" inherit_id="survey.survey_page_print">
+        <xpath expr="//t[@t-call='survey.question_matrix']/.." position="after">
+            <t
+                t-if="question.question_type == 'model'"
+                class="o_survey_print o_survey_comment_container p-0"
+            >
+                <div
+                    class="col-sm-12 o_survey_choice_btn py-1 px-3 w-100 h-100 rounded mb-3"
+                >
+                <span t-out="answer_lines[0].value_model.display_name or None" />
```
carlos-lopez-tecnativa: I got this error when trying to print the survey.
Image

Image


Image

carlos-lopez-tecnativa (changes requested):

survey_question_type_model_selection/static/src/js/survey_form_model.esm.js
```diff
@@ -0,0 +1,312 @@
...
+     * @override
+     */
+    start: function () {
+        const result = this._super.apply(this, arguments);
+
+        // Initialize all inputs multiple times to catch async loads
+        this._initAllInputs();
+        setTimeout(() => this._initAllInputs(), 100);
+        setTimeout(() => this._initAllInputs(), 500);
+        setTimeout(() => this._initAllInputs(), 1000);
```
carlos-lopez-tecnativa: I don't understand why we do the same thing with different intervals. Could you add a better comment or docstring?

survey_question_type_model_selection/static/src/js/survey_form_model.esm.js
```diff
@@ -0,0 +1,312 @@
...
+    _createErrorMessage: function (input, data) {
+        let errorMsg = input.parentElement.querySelector(".model-input-error");
+        if (!errorMsg) {
+            errorMsg = document.createElement("div");
+            errorMsg.className = "model-input-error";
+            errorMsg.style.color = "#dc3545";
+            errorMsg.style.fontSize = "0.875rem";
+            errorMsg.style.marginTop = "0.25rem";
+            errorMsg.style.display = "none";
+            errorMsg.textContent = "Please select a valid option from the list";
```
carlos-lopez-tecnativa: I think this text can be translated.

survey_question_type_model_selection/views/survey_template.xml
eduezerouali-tecnativa: This already fix, now it manages if it has no answer.

survey_question_type_model_selection/static/src/js/survey_form_model.esm.js
eduezerouali-tecnativa: Not really need it, it is done in `this._setupObserver();` ,we need init the selection. becuse when DOM is display first time it is not load. I don't know if there is a more practical way of implementing it.

survey_question_type_model_selection/static/src/js/survey_form_model.esm.js
eduezerouali-tecnativa: Now it can be translated

eduezerouali-tecnativa:
@carlos-lopez-tecnativa thanks to take the time for review this PR. I did attend your comments. It is ready for review again. Thank you.

survey_question_type_model_selection/static/src/js/survey_form_patch.esm.js
```diff
@@ -0,0 +1,31 @@
...
+    const maxAttempts = 50;
+    const interval = setInterval(() => {
+        attempts++;
+        if (applySurveyPatch()) {
+            clearInterval(interval);
+        } else if (attempts >= maxAttempts) {
+            clearInterval(interval);
+        }
+    }, 100);
+}
```
carlos-lopez-tecnativa: Why? Could you add a docstring to explain the reason for this code? It is not clear to me.
I imagine the DOM is not ready yet, but what is the actual reason?

survey_question_type_model_selection/static/src/js/survey_form_model.esm.js
```diff
@@ -0,0 +1,304 @@
+/** @odoo-module **/
+import {_t} from "@web/core/l10n/translation";
+import publicWidget from "@web/legacy/js/public/public_widget";
+
+publicWidget.registry.SurveyModelInput = publicWidget.Widget.extend({
+    selector: ".js_surveyform, .o_survey_form, body",
```
carlos-lopez-tecnativa: I’m not sure if this code is loaded correctly. I attached a video where, if I start the survey from the test (or more specifically, if it displays the main page before starting the survey), it works fine. However, when I start the survey from a link from the Share wizard, it does not display the main page. The first question is loaded, but the input does not display the options. I suspect it is not loading correctly. Please take a look.

https://github.com/user-attachments/assets/2d092b3b-30a3-4769-b91b-f9c6c95cb2ff

carlos-lopez-tecnativa (changes requested):

carlos-lopez-tecnativa:
I see the code adds a feature to allow the user to select options, similar to a many2one field, which is nice, but what is the reason for implementing this here? Could an external library be used instead, such as `Choices.js` or `Tom Select`, etc?

pedrobaeza:
For such simple feature, it's better to not add such maintenance burden.

eduezerouali-tecnativa:
Add choices to it. Main changes apply:

- Add choices lib.
- survey_form.esm.js add choices + remove unnecessary code.
- Templates match new select.
- Tours adapted to select.

Changes do not affect following PR https://github.com/OCA/survey/pull/210 https://github.com/OCA/survey/pull/211
Also add @carlos-lopez-tecnativa as contributor as main reviewer. 

please @pilarvargas-tecnativa @carlos-lopez-tecnativa  could you review.

survey_question_type_model_selection/models/survey_question.py
```diff
@@ -0,0 +1,36 @@
...
+class SurveyQuestion(models.Model):
+    _inherit = "survey.question"
+
+    question_type = fields.Selection(selection_add=[("model", "Model selection")])
+    question_model_id = fields.Many2one(
+        string="Applies to",
+        comodel_name="ir.model",
+        default=lambda self: self.env["ir.model"]._get_id("res.partner"),
+        ondelete="cascade",
+    )
```
carlos-lopez-tecnativa, line 18: Please add a domain to prevent selecting models such as `TransientModel` and `AbstractModel`.

Because if a user selects a model like mail.thread, this causes an error in the frontend, and the domain is evaluated incorrectly.

Image

Image

survey_question_type_model_selection/models/survey_question.py
```diff
@@ -0,0 +1,36 @@
...
+import ast
+
+from odoo import fields, models
+
+
+class SurveyQuestion(models.Model):
+    _inherit = "survey.question"
+
+    question_type = fields.Selection(selection_add=[("model", "Model selection")])
+    question_model_id = fields.Many2one(
```
carlos-lopez-tecnativa, line 12: Take into account that if a user does not have permission to read `ir.model` (the group required to grant this access is `Administration / Access Rights`), they will get an error when trying to add a new question.
Image

Image

carlos-lopez-tecnativa (changes requested):
Tested functionally; it works. I left some technical comments to prevent mistakes.

Regarding the new library, it works fine, thanks. I didn’t request this as a change; it was just a question. If you decide to use an external library, that’s OK for me. Let’s see what @pedrobaeza thinks about this 😅

pedrobaeza:
Yes, adding a JS library if there are more reasonable options is not the first choice (wordplay with the library name, hehe), and more if there are license incompatibilities (don't know the library license). What is that library bringing?

eduezerouali-tecnativa:
It is a MIT https://github.com/Choices-js/Choices so there is not conflict using it. Is there any reason to add more code to have a custom many2one in the survey instead of using this library?

pedrobaeza:
The reason is to not having to bundle a very heavy library inside the module, but if the ratio gain/pain doesn't worth, let's use the library.

survey_question_type_model_selection/models/survey_question.py
eduezerouali-tecnativa: done

survey_question_type_model_selection/models/survey_question.py
eduezerouali-tecnativa: done

eduezerouali-tecnativa:
@carlos-lopez-tecnativa @pilarvargas-tecnativa could you review please?

carlos-lopez-tecnativa (approved):
LGTM

pilarvargas-tecnativa (approved):

pedrobaeza:
@sbidoul there's no way to get this merged (among others). Please create manually the project.

sbidoul:
Done. Please retry the merge.

PR #211, opened 2025-12-16
[17.0][ADD] survey_contact_generation_model_selection: new module

eduezerouali-tecnativa:
Depends on:
 - [x]  https://github.com/OCA/survey/pull/209

Working on:
 - [x] Test
 
cc @Tecnativa TT59700
ping @pilarvargas-tecnativa @david-banon-tecnativa @pedrobaeza


eduezerouali-tecnativa:
@pilarvargas-tecnativa @carlos-lopez-tecnativa could you review please?

survey_contact_generation_model_selection/models/survey_question.py
```diff
@@ -0,0 +1,20 @@
...
+    @api.depends("question_type")
+    def _compute_allowed_partner_field_domain(self):
+        res = super()._compute_allowed_partner_field_domain()
+        for record in self:
+            if record.question_type == "model":
+                record.allowed_partner_field_domain = [
+                    ("model", "=", "res.partner"),
+                    ("ttype", "in", ["many2one", "many2many"]),
+                    ("store", "=", True),
+                    ("readonly", "=", False),
```
carlos-lopez-tecnativa, line 20: I think you need to filter based on the relation field so that it matches the model defined in Applies to (`question_model_id`).

In my example, I can configure the model as `res.country`, but in `res_partner_field` I specify that the value must be taken from the `category_id` field, whose related model is `res.partner.category`. As a result, the configured model and the field's related model are different, so the IDs do not match.

Image

Image

Image

carlos-lopez-tecnativa (changes requested):

survey_contact_generation_model_selection/models/survey_question.py
eduezerouali-tecnativa: We could add`("relation", "=", record.question_model_id.model)` and `question_model_id` to the depends. And a constrain to point out if the do not match. What do you think?

eduezerouali-tecnativa: Done!

carlos-lopez-tecnativa (approved):
LGTM, please check the CI

pilarvargas-tecnativa (approved):
LGTM 👍 

Please, do a rebase

PR #212, opened 2026-01-24
[18.0][ADD] survey_question_list_all: New module

Daniel15:
# Problem
Surveys can get quite long if they're used for collecting data from clients. By default, one2many and many2many lists in Odoo only show 40 items, requiring pagination after that. Additionally, after adding a new question, it goes back to the start of the list rather than staying on the right page, making it frustrating to quickly iterate on new questions.

# This PR
This PR adds a new module that updates the survey form view to show all questions, not just 40. Technically it changes the page size to 10000, but that should be sufficient for anything.

Before:
Screenshot_20260123_210225

After:
Screenshot_20260123_210159

This is the first time I've created an Odoo module, so please let me know if there's any issues.


Daniel15:
For now this is in my own repo https://github.com/Daniel15/odoo-modules. When I get some free time, I'll create a new PR with the current state of the repo, as I've changed the module a bit.

PR #213, opened 2026-01-24
[18.0][ADD] survey_debrand: New module

Daniel15:
Adds a basic module that removes the "Powered by Odoo" branding from survey pages: 
image

I named it `survey_debrand` to keep the naming similar to `mail_debrand` which does a similar thing but for emails.


flotho (approved):
successfully tested on runboat
LGTM,
Thanks

Daniel15:
For now this is in my own repo https://github.com/Daniel15/odoo-modules. When I get some free time, I'll create a new PR with the current state of the repo, as I've changed the module a bit.

PR #214, opened 2026-02-04
[16.0][FIX] survey_certification_branding

edescalona:
@BinhexTeam

When a portal user registers and completes a certification, the following errors occur:

- When the text in the "Company Name" field in the "Certification Mark" section and the survey name are too long, they overlap with the date and image values ​​for the "Certification Mark." Therefore, limiting these to 100 and 180 characters, respectively, is recommended.
- The image configured in the "Certification Mark" section of the survey is not displayed.

This change request resolves both of these issues.

Survey title:
image

Company name (Certification branding)
image

[FIX] Before (Certification Modern)
image

[FIX] Before (Certification Classic)
image

After (Certification Modern)
image

After (Certification Classic)
image

### DIFF
--- a/survey_certification_branding/__manifest__.py
+++ b/survey_certification_branding/__manifest__.py
@@ -16,4 +16,9 @@
         "views/survey_survey_views.xml",
         "report/survey_report_templates.xml",
     ],
+    "assets": {
+        "web.report_assets_common": [
+            "survey_certification_branding/static/src/scss/survey_reports.scss",
+        ],
+    },
 }
--- a/survey_certification_branding/models/survey_survey.py
+++ b/survey_certification_branding/models/survey_survey.py
@@ -7,6 +7,7 @@ from odoo import api, fields, models
 class SurveySurvey(models.Model):
     _inherit = "survey.survey"
 
+    title = fields.Char(size=180)
     certification_company_name = fields.Char(
         "Certification - Company Name",
         help=(
@@ -15,6 +16,7 @@ class SurveySurvey(models.Model):
         ),
         compute="_compute_certification_branding_fields",
         store=True,
+        size=100,
         readonly=False,
     )
 
--- a/survey_certification_branding/readme/CONTRIBUTORS.md
+++ b/survey_certification_branding/readme/CONTRIBUTORS.md
@@ -1,1 +1,2 @@
 - Rolando Pérez Rebollo <r.perez@binhex.cloud> (https://binhex.cloud)
+- Edilio Escalona Almira <e.escalona@binhex.cloud> (https://binhex.cloud)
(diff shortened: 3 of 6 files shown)


edescalona:
ping @rrebollo

survey_certification_branding/README.rst
```diff
@@ -17,7 +13,7 @@ Survey Certification Branding
 .. |badge1| image:: https://img.shields.io/badge/maturity-Beta-yellow.png
     :target: https://odoo-community.org/page/development-status
     :alt: Beta
-.. |badge2| image:: https://img.shields.io/badge/license-AGPL--3-blue.png
+.. |badge2| image:: https://img.shields.io/badge/licence-AGPL--3-blue.png
```
atgalvez08, line 16: The license badge was changed from license to licence. OCA consistently uses license in badges, and the README is autogenerated, so this change should be reverted or applied in the readme source instead.

atgalvez08 (changes requested):

survey_certification_branding/README.rst
edescalona: Hi @atgalvez08 , that's how the pre-commit generates it; perhaps it's a missing configuration. @pedrobaeza , could you comment on this? Thanks

pedrobaeza: This is because the README template that uses ocabot is not the repository one, but a custom fixed one, provoking this problem. @sbidoul is finally the template included in the latest oca-addons-repo-template?

edescalona:
ping @atgalvez08 @rrebollo

rrebollo (commented):
@edescalona Good catch! Did you consider handling the text overflow with CSS (SCSS)?

edescalona:
Hi @rrebollo , yes we looked into that case with CSS, but it's not the best approach here. The goal is to avoid very long content and the certificate stretching across two pages.

edescalona:
Hi @sbidoul , could you please review @pedrobaeza  comment or tell us how to proceed? Thanks

pedrobaeza:
I think that if you update copier template to the latest one, now it's included.

edescalona:
ping @pedrobaeza @sbidoul

rrebollo (commented):
Friend, you need to remove the `index.html` and `README.rst` files from the commit—not from the filesystem. Also, the `pre-commit-config.yaml` approach is not the right way to go.

edescalona:
Hi @rrebollo @pedrobaeza @atgalvez08 , please review this when you can; the incorrect files have been removed.

rrebollo (approved):
Code Review. LGTM!

atgalvez08 (approved):
Code Review. LGTM!

PR #215, opened 2026-02-17
[17.0][FIX] survey_sale_generation: Make the tour selector more resilient

pilarvargas-tecnativa:
If any module modifies the survey completion message, it may cause an error when searching for that text. To avoid this situation, a more resilient selector is added, searching for a specific class on the survey completion page.

@Tecnativa

@pedrobaeza @CarlosRoca13 please review


pedrobaeza (approved):
This can be fast-tracked.

/ocabot merge patch

PR #224, opened 2026-05-28
18.0 mig survey answer generation

layadelcarmen:


layadelcarmen:
@benbrich Check this out when you have a chance

layadelcarmen:
While testing this migration locally, I found that pre-filled answers on the linked survey (next_survey_id) could not be overwritten for any question type other than save_as_email/save_as_nickname.

Root cause: Odoo core's /survey/submit controller only passes overwrite_existing=True when the survey allows going back or when the question is flagged save_as_nickname/save_as_email ([survey/controllers/main.py:538](https://github.com/odoo/odoo/blob/18.0/addons/survey/controllers/main.py#L538)). For every other question type, resubmitting a pre-filled answer on the linked survey (Survey B) raised UserError: "This answer cannot be overwritten" — which also fails silently on the frontend, since /survey/submit is a type="json" route and survey_form.js doesn't catch the rejection, so nothing is shown to the end user; the error only surfaces in the server logs.

Since the whole point of this module is letting users review/edit pre-filled answers before submitting Survey B (and comparing the diff against the origin), this made the "review and change" workflow unusable for anything but email/nickname questions — same behavior reproduced on the v17 demo, so it's not migration-introduced, just previously untested.

Fix: override _save_lines to force overwrite_existing=True whenever the input has an origin_input_id (i.e. it's the linked/generated survey), regardless of question type. Added a regression test (test_overwrite_existing_forced_for_linked_answer) that calls _save_lines(..., overwrite_existing=False) directly — mirroring what the controller does — to make sure this doesn't regress silently again, since the existing diff-related tests mutate the field directly and don't exercise this path.

layadelcarmen:
@benbrich @chandni299  this one's ready for review. Would appreciate a look when you have time.

PR #234, opened 2026-07-14
[17.0][IMP] survey_contact_generation: add posibility to create more than one partner per survey

eduezerouali-tecnativa:
cc @Tecnativa TT63375
ping @pedrobaeza @pilarvargas-tecnativa


survey_contact_generation/models/survey_question.py
```diff
@@ -13,6 +13,26 @@ class SurveyQuestion(models.Model):
...
     )
+    survey_question_node_id = fields.Many2one(
+        comodel_name="survey.question",
+        string="Node",
+        help="Question opening the contact this one relates to. A question "
+        "mapped to the name opens a contact of its own, and hangs it from the "
+        "contact of this node. Any other question fills in a field of the "
+        "contact this node opened.",
+    )
+    # Modifiers get a many2one as its bare id, so the field name has to be
```
pedrobaeza: I don't get this comment

survey_contact_generation/views/survey_question_views.xml
```diff
@@ -11,6 +11,17 @@
                         options="{'no_create': True, 'no_open': True}"
                         domain="allowed_partner_field_domain"
                     />
+                    <field
+                        name="survey_question_node_id"
+                        options="{'no_create': True, 'no_open': True}"
+                        invisible="not res_partner_field"
```
pedrobaeza, line 17: Shouldn't this be mandatory if `res_partner_field`?

survey_contact_generation/models/survey_question.py
eduezerouali-tecnativa: I was trying to explain why that field is necessary.

pedrobaeza: Maybe you can rephrase it?

survey_contact_generation/views/survey_question_views.xml
eduezerouali-tecnativa: Not really as we did not modify old behavior just added the possibility of creating multiple contacts. So in case you only want to create one contact that field should be empty.

pedrobaeza: Oh, true.

pedrobaeza (approved):

survey_contact_generation/models/survey_question.py
eduezerouali-tecnativa: Done, I hope it is more clear now.

adasatorres-tecnativa (approved):
Code and functional review

PR #240, opened 2026-08-06
[19.0][MIG] partner_survey: Migration to version 19.0

eduezerouali-tecnativa:
cc @Tecnativa TT63998
ping @pilarvargas-tecnativa @adasatorres could you review please?


partner_survey/models/res_partner.py
```diff
@@ -0,0 +1,71 @@
...
+    @api.depends("is_company")
+    def _compute_surveys_count(self):
+        read_group_res = (
+            self.env["survey.user_input"]
+            .sudo()
+            ._read_group(
+                [("partner_id", "in", self.ids)], ["partner_id"], ["partner_id:count"]
+            )
+        )
+        data = {res["partner_id"][0]: res["partner_id_count"] for res in read_group_res}
```
adasatorres-tecnativa: I have checked on Runboat and an error occurs when opening a contact that has an associated survey, or when the survey belongs to one of its child contacts.

Image

In this version, _read_group returns a list of tuples containing a singleton record and its count.

```suggestion
        data = {res[0].id: res[1] for res in read_group_res}
```

adasatorres-tecnativa (changes requested):

partner_survey/models/res_partner.py
eduezerouali-tecnativa: Done. Thank you!

PR #236, opened 2026-08-06
[19.0][MIG] survey_legal: Migration to version 19.0

adasatorres-tecnativa:
Standard Migration
@Tecnativa TT64013

PR #243, opened 2026-08-06
[19.0][IMP]* copier update to v1.44

eduezerouali-tecnativa:
cc @Tecnativa
ping @pedrobaeza @pilarvargas-tecnativa 
This is to avoid `chrome (zombie)` warning in https://github.com/OCA/survey/pull/236 and https://github.com/OCA/survey/actions/runs/31104156328/job/92624887987?pr=241

### DIFF
--- a/.copier-answers.yml
+++ b/.copier-answers.yml
@@ -1,8 +1,7 @@
 # Do NOT update manually; changes here will be overwritten by Copier
-_commit: v1.35
+_commit: v1.44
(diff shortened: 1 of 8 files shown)


pedrobaeza:
CI is red

eduezerouali-tecnativa:
Well, CI is failing because there isn't any module to test. How do you think we should proceed?

pedrobaeza:
OK, let's merge it then and it will be fixed with the first one. This error should be handled at tools level. cc @sbidoul

PR #247, opened 2026-08-12
[19.0][MIG] survey_sale_generation: Migration to version 19.0

adasatorres-tecnativa:
@Tecnativa TT64020


survey_sale_generation/models/survey_user_input.py
```diff
@@ -81,15 +81,15 @@ def _prepare_quotation_comment(self):
...
-        message = _(
-            "This order has been created from this survey input: "
-            "<a href=# data-oe-model=survey.user_input data-oe-id=%(id)d>%(title)s</a>"
-        ) % {"id": self.id, "title": self.survey_id.title}
-
+        message = self.env._("This order has been created from this survey input: ")
+        message += (
+            "<a href=# data-oe-model=survey.user_input "
+            f"data-oe-id={self.id}>{self.survey_id.title}</a>"
+        )
```
eduezerouali-tecnativa: I think you should keep it as it was with the correct formatting

eduezerouali-tecnativa (changes requested):

survey_sale_generation/models/survey_user_input.py
adasatorres-tecnativa: @eduezerouali-tecnativa  Ruff in the pre-commit hook is asking me to use f-strings instead.

survey_sale_generation/models/survey_user_input.py
```diff
@@ -0,0 +1,150 @@
...
+
+    def _create_quotation_post_process(self):
+        """After creating the quotation send an internal message with practical info"""
+        sale_sudo = self.sale_order_id.sudo()
+        message = self.env._("This order has been created from this survey input: ")
+        message += (
+            "<a href=# data-oe-model=survey.user_input "
+            f"data-oe-id={self.id}>{self.survey_id.title}</a>"
+        )
```
eduezerouali-tecnativa:
```suggestion
        message = self.env._(
            "This order has been created from this survey input: "
            "<a href=# data-oe-model=survey.user_input data-oe-id=%(id)d>%(title)s</a>",
            id=self.id,
            titles=self.survey_id.title,
        )
```
I meant something like this should be valid for the pre-commit and we can keep it as it was.

adasatorres-tecnativa: Sorry, I always forget that they can also be passed as arguments to the call to _.

survey_sale_generation/models/survey_user_input.py
pedrobaeza: There's a method at model level to get this link. Please investigate it and replace it with it.

eduezerouali-tecnativa (approved):

adasatorres-tecnativa:
ping @pilarvargas-tecnativa

survey_sale_generation/tests/test_survey_sale_generation.py
```diff
@@ -11,91 +13,274 @@ def setUp(self):
...
-        Mail Management          $500.00             1.000
-        Support hours            $300.00             3.000
-        """
         expected_lines = {
-            self.support_hours: 3,
-            self.gold_service: 3,
-            self.advanced_backup: 1,
-            self.mail_management: 1,
+            self.support_hours: 3.0,
+            self.gold_service: 1.0,
```
pilarvargas-tecnativa: Could you please check why this was 3 in 17.0 while the documented expected result was 1? The test was passing with 3, so we should determine whether there was a bug in the 17.0 logic/configuration or whether the documentation was incorrect. If 1 is indeed the correct behavior in 19.0, it would be good to clarify why this change is valid as part of the migration.

survey_sale_generation/tests/test_survey_sale_generation.py
```diff
@@ -11,91 +13,274 @@ def setUp(self):
...
+            self.advanced_backup: 1.0,
+            self.mail_management: 1.0,
         }
         resulting_lines = {
             line.product_id: line.product_uom_qty
             for line in self.generated_sale.order_line
         }
-        self.assertEqual(resulting_lines, expected_lines)
```
pilarvargas-tecnativa: Why do we need to split the previous dictionary assertion into individual assertions? self.assertEqual(resulting_lines, expected_lines) also ensures that no unexpected sale order lines are generated, so it seems to provide better coverage.

survey_sale_generation/tests/test_survey_sale_generation.py
adasatorres-tecnativa: I don't see any indication in the 17.0 tour that suggests the gold service should be at 3. Although the tests in 17.0 pass perfectly, is it possible that there is a bug in 17.0? 

There is a comment indicating what the sales order used should look like. It states that the quantity for that product should be 1.

pilarvargas-tecnativa: I think my previous comment was misunderstood. I’m not saying that this should necessarily be changed to 1 because the comment says so.

What I mean is that in 17.0 the test was passing expecting 3, while the documented result says 1. Before changing the expected value as part of the migration, we should understand why: whether 17.0 had a bug in the multiplier logic/configuration, or whether the comment was simply wrong and 3 was actually the intended behavior.

If 1 is confirmed to be the correct behavior, then the change is fine, but I think we should first verify why the previous version generated 3.

adasatorres-tecnativa:
There’s actually no issue with the tests. When I was transcribing the demo survey into the test setup, I forgot to add a field to the question about the Gold services. That field indicates whether you want to use a numeric field as a multiplier for the value of that question.

Sorry for any inconvenience caused.

@pilarvargas-tecnativa.

pilarvargas-tecnativa:
> There’s actually no issue with the tests. When I was transcribing the demo survey into the test setup, I forgot to add a field to the question about the Gold services. That field indicates whether you want to use a numeric field as a multiplier for the value of that question.

Thanks for checking and clarifying! The confusion came from the 17.0 test documentation showing 1 while the actual expected behavior was 3. Thanks also for correcting it in 19.0 🙂

survey_sale_generation/tests/test_survey_sale_generation.py
```diff
@@ -11,91 +13,265 @@ def setUp(self):
...
             line.product_id: line.product_uom_qty
             for line in self.generated_sale.order_line
         }
         self.assertEqual(resulting_lines, expected_lines)
         self.assertEqual(self.generated_sale.team_id, self.support_hiring_team)
-        self.assertEqual(
-            self.generated_sale.partner_id, self.env.ref("base.partner_demo_portal")
-        )
-        self.assertEqual(
-            self.generated_sale.sale_order_template_id, self.quotation_template_1
```
pilarvargas-tecnativa: Would it be possible to restore this assertion to keep the default survey quotation template behavior covered?

adasatorres-tecnativa:
@pilarvargas-tecnativa, It should already cover the check you mentioned.

pilarvargas-tecnativa (approved):

PR #252, opened 2026-08-24
[19.0][MIG] survey_contact_generation_model_selection: Migration to version 19.0

adasatorres-tecnativa:
@Tecnativa TT64009


adasatorres-tecnativa:
@eduezerouali-tecnativa @pilarvargas-tecnativa

eduezerouali-tecnativa:
Could you rebase please

adasatorres-tecnativa:
ready @eduezerouali-tecnativa

survey_contact_generation_model_selection/tests/test_contact_generation_model_selection.py
```diff
@@ -1,57 +1,61 @@
...
         cls.survey = cls.env["survey.survey"].create(
             {
                 "title": "Test Question Model",
-                "survey_type": "custom",
                 "questions_layout": "one_page",
                 "access_mode": "public",
-                "access_token": "b137640d-14d4-4748-9ef6-344caaaaaaf",
+                "users_can_go_back": True,
                 "generate_contact": True,
+                "create_parent_contact": True,
```
eduezerouali-tecnativa: This is not necessary as this test does not create the parent.

survey_contact_generation_model_selection/tests/test_contact_generation_model_selection.py
```diff
@@ -1,57 +1,61 @@
...
+from odoo.addons.survey.tests.common import SurveyCase
 
 
 @tagged("-at_install", "post_install")
-class TestContactGenerationModelSelection(BaseCommon, HttpCase):
+class TestContactGenerationModelSelection(SurveyCase, HttpCase):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
+        new_test_user(cls.env, login="test-user", groups="base.group_portal")
```
eduezerouali-tecnativa: You are adding a new user, but it is not used in the tour. Anyways if you do it that way the partner will not be created.

survey_contact_generation_model_selection/__manifest__.py
```diff
@@ -3,17 +3,16 @@
...
     "maintainers": ["eduezerouali-tecnativa"],
     "license": "AGPL-3",
     "depends": ["survey_contact_generation", "survey_question_type_model_selection"],
-    "assest": {
+    "assets": {
         "web.assets_tests": [
             "survey_contact_generation_model_selection/static/tests/test_tour_contact_generation_model_selection.esm.js",
         ],
     },
-    "auto_install": True,
```
eduezerouali-tecnativa: You should keep this as it is a glue module.

eduezerouali-tecnativa (changes requested):
Thank you! Can you add this https://github.com/OCA/survey/pull/257, so the module works as it should.

adasatorres-tecnativa:
ready @eduezerouali-tecnativa

eduezerouali-tecnativa (approved):
LGTM + Tested on runboat

adasatorres-tecnativa:
@pilarvargas-tecnativa could you please review this?

survey_contact_generation_model_selection/__manifest__.py
```diff
@@ -3,14 +3,14 @@
...
-    "version": "17.0.1.0.0",
+    "version": "19.0.1.0.0",
     "category": "Marketing/Survey",
     "website": "https://github.com/OCA/survey",
     "author": "Tecnativa, Odoo Community Association (OCA)",
     "maintainers": ["eduezerouali-tecnativa"],
     "license": "AGPL-3",
     "depends": ["survey_contact_generation", "survey_question_type_model_selection"],
-    "assest": {
+    "assets": {
```
pilarvargas-tecnativa, line 13: The typo seems to originate in version 17, could it be corrected in that version? Ideally, it should be corrected in version 17 and carried over to version 19, but in any case it must be corrected in version 17 as well.

survey_contact_generation_model_selection/static/tests/test_tour_contact_generation_model_selection.esm.js
```diff
@@ -1,37 +1,51 @@
...
+                // Now click input explicitly
+                content: "focus input",
+                trigger: "div.js_question-wrapper:contains('State') div.choices",
+                run: "click",
+            },
+            {
+                // Wait for dropdown to appear
+                content: "wait dropdown",
+                trigger:
+                    "div.js_question-wrapper:contains(State') div.choices option:contains('Murcia (ES)')",
```
pilarvargas-tecnativa: Is this step actually being reached by the tour? The selector contains `:contains(State')`, while the question title is State, so I would expect the tour to timeout here.

pilarvargas-tecnativa (commented):
Please reduce administrative commits

survey_contact_generation_model_selection/__manifest__.py
eduezerouali-tecnativa: This was already fix in 17

pilarvargas-tecnativa: Well, in that case, simply bring the fix commit before the migration commit to preserve the commit history.

survey_contact_generation_model_selection/models/survey_user_input.py
```diff
@@ -0,0 +1,17 @@
...
+
+    def _prepare_partner_vals(self, elegible_inputs):
+        res = super()._prepare_partner_vals(elegible_inputs)
+        for key, value in res.items():
+            if isinstance(value, models.BaseModel):
+                if len(value) == 1:
+                    res[key] = value.id
+                else:
+                    res[key] = [Command.set(value.ids)]
+        return res
```
cristina-hidalgo-tecnativa, line 19:
```suggestion
     def _prepare_partner_vals(self, elegible_inputs):
        res = super()._prepare_partner_vals(elegible_inputs)
        for line in elegible_inputs.filtered(lambda x: x.answer_type == "model"):
            field = line.question_id.res_partner_field
            if not field or field.name not in res:
                continue
            if field.ttype == "many2many":
                res[field.name] = [Command.set(line.value_model.ids)]
            else:
                res[field.name] = line.value_model.id
        return res
```

While testing this locally and on runboat, I submitted a public survey as an anonymous user (no login — the real use case for `survey_contact_generation`, since a logged-in user already has `partner_id` set and skips generation).

Image Image

Runboat:

Image

It works fine when the model question is mapped to a `many2one` contact field (e.g. `state_id`), but if it's mapped to a `many2many` one the submit fails and the survey ends in an `RPC_ERROR`, so it can't be completed and no contact is created.

Image

It comes from `_prepare_partner_vals` choosing between `many2one` and `many2many` based on `len(value)`. Since a model question is always single-select, `len(value)` is always `1`, so the `Command.set` branch never runs and a `many2many` field ends up getting a bare int, which `create()` rejects. And a `many2many` field is a valid choice here — `_compute_allowed_partner_field_domain` allows it explicitly.

I think it's cleaner to look at the field type instead.

cristina-hidalgo-tecnativa (changes requested):

survey_contact_generation_model_selection/models/survey_user_input.py
eduezerouali-tecnativa: Thanks for pointing this out! Fix is here https://github.com/OCA/survey/pull/258, so @adasatorres-tecnativa could you include that commit in this migration.

survey_contact_generation_model_selection/static/tests/test_tour_contact_generation_model_selection.esm.js
adasatorres-tecnativa: For me, the question appears in English, and when I run the tests, everything works perfectly. It might be because the question title is a translatable field. I understand that if the test is run with a language other than English, the tour will fail from the very first step.

adasatorres-tecnativa:
ready @eduezerouali-tecnativa @cristina-hidalgo-tecnativa

survey_contact_generation_model_selection/static/tests/test_tour_contact_generation_model_selection.esm.js
pilarvargas-tecnativa: Yes, the translation could be another issue, but I was referring specifically to the missing quote in :contains(State'). Shouldn't it be :contains('State')? I'm surprised the test passes with the current selector.

cristina-hidalgo-tecnativa (approved):
Code reviewed and tested locally and on runboat. LGTM

Not blocking, but I'd suggest adding a test covering the m2o/m2m mapping that was just fixed.

survey_contact_generation_model_selection/static/tests/test_tour_contact_generation_model_selection.esm.js
pedrobaeza: @adasatorres-tecnativa have you checked this?

pedrobaeza: ping @adasatorres-tecnativa

adasatorres-tecnativa: @pedrobaeza yes, I’ve reviewed it, and I’ve also added and corrected that step in the test because, when adding the ', it wasn’t working correctly.

PR #253, opened 2026-08-24
[17.0][FIX] survey_contact_generation_model_selection: fix assets key on __manifest__.py

adasatorres-tecnativa:
@Tecnativa TT64009
@pedrobaeza @eduezerouali-tecnativa


adasatorres-tecnativa:
The tour name was causing problems because it was the same as another tour. I’ve also changed some relational fields that were being sent with hardcoded values instead of using their reference.

survey_contact_generation_model_selection/static/tests/test_tour_contact_generation_model_selection.esm.js
```diff
@@ -1,37 +1,41 @@
...
-        },
-        {
-            content: "Click Submit",
-            trigger: "button[value='finish'].btn-secondary",
-        },
-    ],
-});
+registry
+    .category("web_tour.tours")
+    .add("test_survey_generation_question_model_selection", {
```
pedrobaeza: Let's be exact:

```suggestion
    .add("test_survey_contact_generation_model_selection", {
```

pedrobaeza (approved):
Thanks for fixing the test.

/ocabot merge patch

PR #254, opened 2026-08-25
[19.0][MIG] survey_next_survey_update_partner: Migration to version 19.0

adasatorres-tecnativa:
@Tecnativa TT64015


survey_next_survey_update_partner/static/tests/survey_contact_update_tour.esm.js
```diff
@@ -0,0 +1,108 @@
+import {registry} from "@web/core/registry";
+
+registry.category("web_tour.tours").add("test_survey_contact_update", {
+    test: true,
```
eduezerouali-tecnativa:
```suggestion
```

eduezerouali-tecnativa:
@adasatorres-tecnativa can you rebase please

adasatorres-tecnativa:
@eduezerouali-tecnativa ready

survey_next_survey_update_partner/demo/survey_update_generated_partner_demo.xml
```diff
@@ -43,13 +43,6 @@
         <field name="question_type">numerical_box</field>
         <field name="res_partner_field" ref="base.field_res_partner__color" />
     </record>
-    <record model="survey.question" id="survey_next_contact_q4">
```
eduezerouali-tecnativa: Why is this removed?

eduezerouali-tecnativa (changes requested):

survey_next_survey_update_partner/demo/survey_update_generated_partner_demo.xml
adasatorres-tecnativa: The field it uses from res.partner does not exist.

cristina-hidalgo-tecnativa (commented):
Tested locally and code-reviewed. 

Migration looks fine and the test suite is green, but the module doesn't work end-to-end on 19.0 — and it's a `survey_answer_generation` bug, not this PR's.

Chain two surveys, map a single/multiple choice question between them (Country, Tags…), and answer them anonymously (private window / no login — otherwise the answer already has a `partner_id` and no contact is generated). Submitting the second survey crashes:

Image

[_compute_diff_with_origin](https://github.com/OCA/survey/blob/17aec3d0515657d656cc261253ed4016e8de69ea/survey_answer_generation/models/survey_user_input.py#L113-L130) doesn't `str()`-coerce the values in the else branch, so a linked choice line with no `suggested_answer_id` yet (prefilled question, not answered while navigating between pages) gives `current_value = False` and `get_diff()` blows up on `False.splitlines()`. Since `diff_with_origin` is a batched stored compute, that one line kills the recompute for all of them → they stay `False` → `_mark_done` never propagates any changed prefilled answer to the partner (only brand-new questions get through).

IMO, it should be fixed first (or in the same series). Worth extending the test here to modify a prefilled choice answer too — the current tour only covers forward `char_box` edits.

survey_next_survey_update_partner/tests/test_next_survey_update_partner.py
```diff
@@ -1,42 +1,207 @@
...
+        cls.next_survey = cls.origin_survey.copy(
+            {"title": "Next contact survey", "question_and_page_ids": questions_vals}
+        )
+        cls.existing_inputs = cls.origin_survey.user_input_ids
+        # Let's links several questions
+        cls.origin_survey.next_survey_id = cls.next_survey
+        survey_contact_q0.next_survey_question_id = survey_next_contact_q0.id
+        survey_next_contact_q_company_name.next_survey_question_id = (
+            survey_contact_q_company_name.id
         )
```
pilarvargas-tecnativa: Could this have been set up the wrong way round? I think the question in the source survey should point to the corresponding question in the next survey, just as it used to.

If it’s set up the other way round, the test might pass, but the question would be treated as a new question without an `origin_input_line`, so the update wouldn’t actually be tested.

survey_next_survey_update_partner/tests/test_next_survey_update_partner.py
```diff
@@ -1,42 +1,207 @@
...
+                                        _create_reference(cls.prospect_category)
+                                    ),
+                                }
+                            ),
+                            Command.create(
+                                {
+                                    "sequence": 3,
+                                    "value": "Employees",
+                                    "res_partner_field_resource_ref": (
+                                        _create_reference(cls.vendor_category)
```
pilarvargas-tecnativa: Shouldn’t it be `cls.employee_category` here? At the moment, `Employees` is referencing the `Vendor` category.

adasatorres-tecnativa:
ping @pilarvargas-tecnativa

pilarvargas-tecnativa (approved):
Code review 👍

pedrobaeza:
ping @eduezerouali-tecnativa

eduezerouali-tecnativa (changes requested):
Code LGTM, but it still not working. Maybe it was not rebase.

Image

Image

adasatorres-tecnativa:
@eduezerouali-tecnativa now.

eduezerouali-tecnativa (approved):
Tested and work fine.

PR #255, opened 2026-08-25
[19.0][MIG] survey_crm_sale_generation: Migration to version 19.0

adasatorres-tecnativa:
@Tecnativa TT64012


eduezerouali-tecnativa:
@adasatorres-tecnativa  Please could you rebase.

adasatorres-tecnativa:
ready @eduezerouali-tecnativa

Andrii9090-tecnativa (approved):
LGTM 👍 
Tested on runboat

survey_crm_sale_generation/tests/test_survey_crm_sale_generation.py
```diff
@@ -0,0 +1,65 @@
...
+        cls.tag_tech_support = _search_or_create(
+            model="crm.tag",
+            domain=[("name", "ilike", "Tech support")],
+            vals={"name": "Tech support"},
+        )
+        cls.tag_survey_leads = _search_or_create(
+            model="crm.tag",
+            domain=[("name", "=", "Survey Leads")],
+            vals={"name": "Survey Leads"},
+        )
```
pilarvargas-tecnativa: Instead of searching for these records by name, could we try to retrieve them by their XML ID using `env.ref(..., raise_if_not_found=False)` and create them only if they do not exist? This way, we avoid relying on the record’s name – which may change due to translations or coincide with other records – and we reuse the specific record defined by the module when it is available.


```suggestion
        def _ref_or_create(xmlid, model, vals):
            record = cls.env.ref(xmlid, raise_if_not_found=False)
            return record or cls.env[model].create(vals)
        cls.tag_tech_support = _ref_or_create(
            "survey_crm_sale_generation.tag_tech_support",
            "crm.tag",
            {"name": "Tech support"},
        )
        cls.tag_survey_leads = _ref_or_create(
            "survey_crm_generation.tag_survey_leads",
            "crm.tag",
            {"name": "Survey Leads"},
        )
```

survey_crm_sale_generation/models/survey_user_input.py
```diff
@@ -0,0 +1,20 @@
...
+
+class SurveyUserInput(models.Model):
+    _inherit = "survey.user_input"
+
+    def _mark_done(self):
+        """Link the generated opportunity and quoation together when the survey
+        is submitted."""
+        res = super()._mark_done()
+        if self.opportunity_id and self.sale_order_id:
+            self.opportunity_id.update(
```
cristina-hidalgo-tecnativa: I'd consider switching `update()` to `write()`: `update()` writes field by field (two `write()` calls here, one per dict key) and is meant for in-memory `onchange` use. A single `self.opportunity_id.write({...})` does the same in one write.

survey_crm_sale_generation/models/survey_user_input.py
```diff
@@ -0,0 +1,20 @@
...
+                {
+                    "expected_revenue": self.sale_order_id.amount_total,
+                    "order_ids": [(6, 0, self.sale_order_id.ids)],
```
cristina-hidalgo-tecnativa:
```suggestion
                    "order_ids": [Command.set(self.sale_order_id.ids)],
```

You can use `Command` here to avoid the legacy (`6, 0, ids`) tuple and make the intent explicit (and stay consistent with the test file, which already imports it).

pilarvargas-tecnativa (approved):

eduezerouali-tecnativa (approved):

PR #259, opened 2026-09-09
[19.0][FIX] survey_answer_generation: problem with get_diff False argument

adasatorres-tecnativa:
@Tecnativa 
@cristina-hidalgo-tecnativa @eduezerouali-tecnativa 

Solution to the issue found in #254

PR #260, opened 2026-09-10
[19.0][FIX] survey_answer_generation: compute_diff_with_origin crash on unanswered suggestion

cristina-hidalgo-tecnativa:
Follow-up to #259 

The else branch of `_compute_diff_with_origin` (suggestion answers) doesn't coerce the values, so when a choice question mapped between two chained surveys is answered in the first one and cleared in the second, `current_value` is `False` and `get_diff()` fails

imagen

@Tecnativa 
@pedrobaeza @adasatorres-tecnativa


adasatorres-tecnativa (approved):
LGTM

pedrobaeza (approved):
/ocabot merge patch

PR #261, opened 2026-09-23
[18.0] [IMP] survey_xlsx: add extension hooks to the XLSX report

mondot:
Extract the per-question header writing, the per-answer processing and a per-user-input post-processing step into dedicated methods called by generate_xlsx_report. This is behavior-preserving and lets downstream modules customize the export without overriding the whole method.

Assisted-By: Claude Opus 4.8 <noreply@anthropic.com>


mondot:
Hi @rousseldenis, here is the PR I talked about earlier with you at OCA Days.

Please let me know if it's enough to be merged :)

